PR #72, opened 2022-11-06
Changes for creation of lists in Neo4j

bnymnsen:
I couldn't create lists from properties in BioCypher's current state. So, I changed some elements to create lists from properties. I added double star mark to join elements of lists then replace this mark with array delimiter in _write.py. With that, I was able to create lists in my properties during admin-import. However, this can be done in _create.py by simply joining lists with selected array delimiter. Another thing I changed is headers. To make headers compatible to list format, I added string[] to header creation so that neo4j can detect the format as a list. Lastly, when I create admin-import-call.sh file after selecting the delimiter as a \t, it writes "    " instead of "\t" which causes error while importing. I manually fixed that issue as well.

### DIFF
--- a/biocypher/_config/test_schema_config.yaml
+++ b/biocypher/_config/test_schema_config.yaml
@@ -31,6 +31,7 @@ protein:
     name: str
     score: float
     taxon: int
+    genes: str[]
 
 microRNA:
   represented_as: node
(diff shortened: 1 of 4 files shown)


bnymnsen:
In addition, I think sensitive marks such as delimiter and array delimiter marks can also be replaced in _create.py

slobentanzer:
2e2490922c5924bc122034dee02ece11f774db13 fixed that for me, should be fine now

PR #145, opened 2023-03-14
fix problems with first level config (like title)

vincentvialard:
key of first level settings are strings, they have no update


slobentanzer (approved):
good addition, thanks :)

PR #160, opened 2023-03-16
Nodes with edges

vincentvialard:
Do not merge yet.

Association names are converted from lower sentencecase to PascalCase; we don't want to do that for edges, but rather leave them as lower snake_case.


vincentvialard:
Added a flag to simple edges to skip PascalCase.

Remark: There is a new unrelated error in the ontology stitching test since merging main again today (get level 7 expected 8).

slobentanzer:
Re the remark: I think this may be due to updates in the ontology, but not sure. I fixed the ontology version for the tests now, but will check later whether I also changed something in the processing. Maybe something happened with the mixins. Thanks for reporting.

PR #188, opened 2023-04-18
add list format to all data types for header creation

bnymnsen:
Small addition to allow list creation for edge headers and for all other data types.

Since it is a small change, I didn't test it with pytest.


slobentanzer (approved):
Looks good, thanks! :)

PR #209, opened 2023-05-26
draft pandas-tutorial in jupyter

dbdimitrov:
Hi @slobentanzer,

I added a simple pandas tutorial in ipynb (it's very much based on the basic tutorial with some minor changes and re-formatting). 

- schemas are printed inline
- Minor text modifications
- Removed the references to each header (I assume duplicates might cause issues, so I leave those to you)
- Used html alerts instead of .md alerts (they look ugly in plain Jupyter, shouldn't matter for the docs page)

It might be best to have this notebook as part of the docs (and maybe remove 01-06 pandas.py scripts, they seem repetitive to me and if this notebook is there, they are not needed imo).

Note that I tagged you in a couple of places with some minor comments.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -92,6 +92,17 @@ optional = false
 python-versions = ">=3.6"
 
 [[package]]
+name = "cffi"
+version = "1.15.1"
+description = "Foreign Function Interface for Python calling C code."
+category = "dev"
+optional = false
+python-versions = "*"
+
+[package.dependencies]
+pycparser = "*"
+
+[[package]]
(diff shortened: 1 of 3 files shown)


slobentanzer:
very nice, will check out ASAP. :)

slobentanzer:
Hi @dbdimitrov,
looks great so far. I guess the most user-friendly way of accessing it would be to have a colab that people can run online, that way we would not need to clone the repo etc. Do you think that's easy enough to implement?

PR #215, opened 2023-06-26
First minimal working version for rdf conversion

supermaxiste:
## Overview
This PR is a first draft to convert BioCypher's nodes and edges into RDF format. The main idea behind this implementation is to offer a minimal solution to convert to RDF on which more elaborate solutions can be built on top. Because of this, the conversion process drops all properties and includes only IDs and labels for nodes and source/target IDs + ID for edges. 

## Implementation details
RDF is considered as a database management system (DBMS). The config file to use this implementation should look as follows:
```YAML
biocypher:
  dbms: rdf

rdf:
  rdf_format: nt 
```
Conversion to RDF is done via `rdflib` which supports `xml`, `n3`, `turtle`, `nt`, `pretty-xml`, `trix`, `trig`, `nquads` and `json-ld` formats.
Since RDF is not a DBMS, I could not reuse the class `_BatchWriter()` but I created a separate class called `_RDFwriter()`. This class still uses some functions from `_BatchWriter()` and has also some RDF-specific ones. I used this approach to follow the general structure of `_write.py`to call a `writer` function, but there might be other approaches, so I'm open to suggestions.

## Major missing component 

This draft doesn't access the underlying ontology yet, meaning that there are no prefixes in the resulting RDF files. I was planning on adding the `OntologyAdapter` class as a dependency in `_write.py`, but then I would also need to add it to the `init` of both `_BatchWriter()` and `_RDFwriter()`, even though it's used just by `_RDFwriter()`. I'd love to hear other ideas for this as well.

## Minor fixes needed

- File extensions need to be adapted based on the specified format. Currently using `turtle` lead to a file called `file.turtle` instead of `file.ttl`. 
- Some file formats have specific requirements that I haven't fully investigated yet
- No unit tests exist for the component yet, could be added as well for future RDF development

### DIFF
--- a/biocypher/_config/biocypher_config.yaml
+++ b/biocypher/_config/biocypher_config.yaml
@@ -110,4 +110,6 @@ postgresql:
   # import_call_bin_prefix: '' # path to "psql"
   # import_call_file_prefix: '/path/to/files'
 
-  
+rdf:
+  ### RDF configuration ###
+  rdf_format: turtle
--- a/biocypher/_core.py
+++ b/biocypher/_core.py
@@ -33,7 +33,7 @@ from ._deduplicate import Deduplicator
 
 __all__ = ['BioCypher']
 
-SUPPORTED_DBMS = ['neo4j', 'postgresql']
+SUPPORTED_DBMS = ['neo4j', 'postgresql', 'rdf']
 
 REQUIRED_CONFIG = [
     'dbms',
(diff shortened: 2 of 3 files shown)


supermaxiste:
Hello @slobentanzer,
thank you for the suggestions and it makes sense to look into PheKnowLator for a converter. 
As for the unit tests, I also agree on prioritizing them.
I'm unsure if I can find time this month to push this PR, but I'll try to find time in September. If you happen to get useful insights before then, feel free to share 👍

slobentanzer:
Hi @supermaxiste,
no worries, we will address this as time allows. :)

Issue #259, opened 2023-08-23
Auto-publish to PyPI


slobentanzer:
review comment: You could publish the package to PyPI via a workflow as well. See [our taxpasta workflow](https://github.com/taxprofiler/taxpasta/blob/dev/.github/workflows/main.yml) for an example of how to do so.

nilskre:
Before adding this feature it makes sense to review the whole CI/CD architecture and specify what the targeted CI/CD architecture is. Therefore here is a proposal for a branching model with CI/CD (based on Gitflow and Trunk-based workflow)

Branches:
- main: contains always the code of the currently released version
- dev: contains the current development state
- feature branches: development of new features
Feature branches are merged frequently into dev and by merging the dev branch into the main branch a new release is created.

On every pushed commit:
- run tests
- ensure code quality

Create new release:
- increase version with bumpversion and commit and push on dev
- create PR to merge dev into main
On every commit to the main branch -> create a new release (triggers CI/CD):
- run tests
- update test coverage
- check code quality
- create release
- publish to PyPi
- update docs

Changes compared to the current state (ToDos):
- Branching model with main, dev and feature branches (dev branch exists in the repo, but is not actively used at the moment)
- add code quality check
- run tests and code quality CI pipeline for every commit (currently only executed for pushes on the main branch)
- create release as part of the CI/CD pipeline

Feedback for this proposal is welcome. If this proposal is alright like this, I can work on a PR for this.

slobentanzer:
Great work making a plan, and I agree that we need to set a target.

My main comment is that I am not a fan of long-running dev branches. The dev branch in the current repo is a good example of why I am not a fan; it contains a lot of work that currently goes unused because it was all written in one go and without tests and failing existing tests. I keep it around for reference and to maybe cherry-pick some of this work, but I haven't gotten around to it. In addition, there were other, parallel developments on other features that made some of these changes redundant or conflicting. As such, I currently work using feature branches exclusively.

There may be a good reason for having a "main `dev` branch", but as far as I understand, this is a philosophical question to some extent. What would be the benefits of having a dev branch? Would we do "pre-releases"? Could we also do the rest of your proposal without the `dev` branch?

nilskre:
I would have not expected an answer this week (:

Yes indeed, this is a philosophical question with pros and cons for both. Without an extra dev branch everything is much simpler and there is not that much overhead.

[...]

Issue #257, opened 2023-08-23
Fix postgres container version


slobentanzer:
review comment: The CI workflow definition creates a postgres container with the latest tag. Are you sure that you don't want to depend on a specific version as you do for neo4j?

Issue #252, opened 2023-08-23
Execute code in documentation to identify outdated function calls


slobentanzer:
The basic tutorial mentions a method `bc.log_missing_bl_types()     # show input` unaccounted for in the schema. This method no longer exists on the BioCypher class. We recommend to execute code in documentation in order to discover such problems quickly.

nilskre:
Used doctest to fix this. 
[Official docs](https://sphinx-tutorial.readthedocs.io/step-3/#testing-your-code)
[Helpful link for the Python syntax](https://mitiq.readthedocs.io/en/stable/contributing_docs.html#run-the-code-with-doctest)

To run all tests within the documentation: `poetry run make doctest`  
This will be integrated into the CI/CD pipeline when working on #259

Issue #251, opened 2023-08-23
Correct tutorial imports


slobentanzer:
review comment:

The basic tutorial suggests to run poetry run python tutorial/01__basic_import.py. This cannot currently work as the line `from tutorial.data_generator import Protein` will fail. The tutorial directory will require an __init__.py module and it would need to be on the PYTHONPATH. Alternatively, users could be required to change the working directory into tutorial and the scripts are adjusted to import the data_generator directly.

slobentanzer:
For me this similarly works, which is no surprise, since I am the one who wrote it. ;)

This was reported by @Midnighter; would you be able to comment directly?

Midnighter:
It might be that I did a simple `python tutorial/01__basic_import.py` rather than a `poetry run`. Can't fully remember. I'll try to check tomorrow.

slobentanzer:
@Midnighter did you find some time to check?

Midnighter:
Sorry, this dropped off my radar for a while.

I can confirm that pulling the latest `main` branch, then `poetry install`, followed by `poetry run ...` works as expected.

slobentanzer:
Great, thanks for the feedback! Closing.

PR #270, opened 2023-09-28
Py open sci review

nilskre:
Fixed issues:
- #251 
- #252 
- #246 
- #195 
- #248 
- #249 
- #250 
- #258 
- #257 

Started working on #259 
Not yet finished and tested, because the trusted publisher mechanism can not be tested from a fork.
The rest of the PyOpenSci issues will be addressed within the next pull request.


.github/workflows/release.yaml
```diff
@@ -1,17 +1,17 @@
-name: Docs build
-on:
--   push
+name: Release
+
+on: [push]
 
 jobs:
-    build:
+    build_and_deploy_docs:
```
slobentanzer, line 6: @nilskre didn't we want to build the docs always, not just on a new release?

slobentanzer:
@nilskre why do the macos runners cancel?

.github/workflows/release.yaml
slobentanzer: maybe a separate yaml for that would be easier to understand / maintain.

nilskre: Yes, we can also build the docs always. I think for the deployment of the docs/artifact there are still some changes to be done within the next PR. So we could have a chat about this and then include the changes in the next PR. Would this be alright?

nilskre: > maybe a separate yaml for that would be easier to understand / maintain.

Yes, I totally agree. So then three yaml files:
- ci_cd.yaml: test and code quality
- docs.yaml: build (and publish?) docs
- release.yaml: build and publish artifact

Are the names alright like this or could we improve understanding by other names?

slobentanzer: > include the changes in the next PR. 

Sure, as long as all checks pass in this PR. I am now thinking that it would probably be easier to check the deploy functionality if we directed this PR at a feature branch (like `pyopensci`) instead of `main` directly. I have redirected the PR; we can merge it with lower risk.

slobentanzer: > Are the names alright

Those are fine, and yes, publishing should also happen in the docs workflow.

slobentanzer:
> the CICD workflow cannot push the new code coverage badge to the coverage branch in the original repo

yes. to be safe, i made a test PR, and the current workflow yaml has no issues pushing to the coverage branch. you could try removing some parts of the workflow iteratively to figure out what causes the permission error. the settings should be fine, the bot is allowed to read and write to branches.

slobentanzer:
Relatedly, I am not sure we should push to the coverage branch on PRs at all. We can still compute, but it is not necessary to update the coverage from all PR CI runs. This should only be done on the `main` branch (meaning when merging something into `main`). Updating this would prevent this step from running in the PR, and thus prevent failure. We still need to make sure it will run on `main`, of course.

nilskre:
Alright, then within the original repo, there are no problems. 
I have also checked to create a PR on my forked repo. All checks are also successful ([PR](https://github.com/nilskre/biocypher/pull/1)).

So I would assume, that the problems arise from creating a PR from a forked repo. I have not yet totally understood the problem. One related discussion about this problem is [here](https://github.com/ad-m/github-push-action/issues/96). So maybe it is related to the "Workflow permissions" of the Actions settings?

nilskre:
> Relatedly, I am not sure we should push to the coverage branch on PRs at all. We can still compute, but it is not necessary to update the coverage from all PR CI runs. This should only be done on the `main` branch (meaning when merging something into `main`). Updating this would prevent this step from running in the PR, and thus prevent failure. We still need to make sure it will run on `main`, of course.

Yes, this makes sense. So we need something like a hook (to update the coverage branch), which is run after merging something into main. I am sure this is possible through Github Actions. I will have a look at it.

nilskre:
Regarding the test coverage config:

tl;dr advantages of using [codecov](https://about.codecov.io/) for managing test coverage:
- Test coverage Dashboard
- PR comments
- free for Open Source projects

Explanation:
Currently, the test coverage is evaluated and the coverage badge is updated, but for new PRs, the change of test coverage is not considered.

In the aimed development scenario of BioCypher, where lots of different persons are developing BioCypher it would be convenient to have specific quality standards enforced for new PRs:
- Done: code quality checks
- Done: running the test suite
- Todo: monitor (and enforce?) a certain level of test coverage for new PRs

This could be achieved by using [codecov](https://about.codecov.io/). 
The usage of codecov is also recommended by the [pyOpenSci Guide](https://www.pyopensci.org/python-package-guide/ci-and-testing/intro.html) ([current proposal](https://github.com/lwasser/python-package-guide/blob/tests/ci-tests-data/code-cov.md)).

What do you think? Would it be an option to use codecov to monitor the test coverage?

slobentanzer:
> Would it be an option to use codecov to monitor the test coverage?

Thanks for the summary. I looked this up briefly when I added the badge, and I decided to go with the "low tech" solution because it has less dependencies. We can consider codecov at a later time (not for this PR), you are welcome to open a dedicated issue for it. :)

There we could have a discussion about what exactly would be gained from using codecov; I am pretty sure that "enforcing" a specific coverage could be done in the low tech way as well. In addition, I don't think enforcing coverage is necessarily the best way to ensure good quality code.

nilskre:
Yes, always good to follow the KISS principle and keep it simple.
I have created an issue for this #272 , so we can discuss this later. (:

.github/workflows/release.yaml
nilskre: To clarify the build and deployment of the docs:
- build the docs -> on every commit

Regarding deployment, I could imagine three scenarios:
[current] 1) Deploy the docs on every tag on the main branch and have one version of the docs: stable (state of the latest tag)
2) Deploy the docs on every commit on the main branch and have two versions of the docs: stable (state of the latest tag) and latest (latest commit on the main branch).
3) Deploy the docs on every commit on the main branch and have multiple versions of the docs: vX.X.X (docs for every version/tag), stable (state of the latest tag), and latest (latest commit on the main branch).

For now, I think option 1 is totally sufficient. But I could imagine, that possibility 2 or 3 could maybe be something for a future improvement?

So for now continue with option 1 and create an issue to discuss options 2 and 3. Would this be alright for you?

.github/workflows/ci_cd.yaml
```diff
@@ -1,116 +1,155 @@
...
-        python-version: '3.9'
-
+      uses: actions/checkout@v4
     - name: Install and configure Poetry
       uses: snok/install-poetry@v1
       with:
         version: 1.5.1
-        virtualenvs-create: false
-
+        virtualenvs-create: true
```
slobentanzer, line 26: @nilskre why do we switch to using venvs?

slobentanzer (approved):

.github/workflows/release.yaml
slobentanzer: Yes, I would keep it simple for now.

nilskre: Great, then this is something to keep in mind when reaching the first major release of BioCypher.

.github/workflows/ci_cd.yaml
slobentanzer: Not particularly, just to keep it simple. These are good reasons, I just wanted to know why. :)

.github/workflows/ci_cd.yaml
slobentanzer, line 1: @nilskre is there an option to use test.pypyi.org in the pyopensci branch to be able to test the release process first?

nilskre: Yes, I have adapted the (former release.yaml) now publish.yaml to upload to test.pypi.org whenever a tag is set. After merging into the pyopensci branch we could set some tag there and test if the upload to test.pypi.org is successful. Is this a suitable procedure for testing?

slobentanzer: yes, that was the basic idea. as soon as the CI passes, I will merge :)

nilskre: Alright, perfect. Now everything is green (:

Issue #253, opened 2023-08-23
Lorem ipsum instead of random letters in data generator


slobentanzer:
review comment: We feel that it would improve the examples, if the data generator created
description fields with actual text (like lorem ipsum), rather than just
random letters.

slobentanzer:
@nilskre maybe look at this dataset: https://ogb.stanford.edu/docs/linkprop/#ogbl-biokg

nilskre:
After looking into this in detail, some thoughts came up.

Specific comments for the dataset:
- The dataset does not contain node features. So there only exists the id of the Protein and a mapping table to map the ids to the underlying Protein ID within a Protein database. Thus, the dataset does not contain the data we need to build up the tutorial (in the tutorial the Protein sequence is used).

General thoughts:
- For new users the tutorial is one of the first touchpoints with BioCypher. 
- On the one hand the usage of a real dataset in the tutorial makes this introduction more realistic.
- On the other hand the increased technical complexity could lead to new unknown errors when running the tutorial (e.g. the download of the dataset already took 5 mins). In the worst case, this could scare off potential new users.

Suggestion:
=> Leave the tutorial as it is, because the core concepts can be easily explained in this toy example scenario.
To improve the readability two suggestions:
- Decrease the randomly generated sequence length to improve readability.
- some real text as description field (to impede the confusion with the Portein sequence).

slobentanzer:
I agree; the tutorial should remain very simple and fast.

The dataset could be a nice foundation for building a more realistic advanced demo (like pole but biomed-related), but that shouldn't be the focus of the basic tutorial.

Issue #226, opened 2023-08-07
Log to disk setting


slobentanzer:
Allow the user to turn off the writing of the log file to disk (in the biocypher config YAML).

PR #283, opened 2023-10-30
Pyopensci really last fixes

nilskre:
Work on the remaining PyOpenSci review tasks:
- #226 
- #253

### DIFF
--- a/biocypher/_logger.py
+++ b/biocypher/_logger.py
@@ -60,7 +60,7 @@ def get_logger(name: str = "biocypher") -> logging.Logger:
         now = datetime.now()
         date_time = now.strftime("%Y%m%d-%H%M%S")
 
-        log_to_disk = _config.config("biocypher").get("log_to_disk", True)
+        log_to_disk = _config.config("biocypher").get("log_to_disk")
 
         if log_to_disk:
             logdir = (
--- a/biocypher/_logger.py
+++ b/biocypher/_logger.py
@@ -60,11 +60,27 @@ def get_logger(name: str = "biocypher") -> logging.Logger:
         now = datetime.now()
         date_time = now.strftime("%Y%m%d-%H%M%S")
 
-        logdir = (
-            _config.config("biocypher").get("log_directory") or "biocypher-log"
-        )
-        os.makedirs(logdir, exist_ok=True)
-        logfile = os.path.join(logdir, f"biocypher-{date_time}.log")
+        log_to_disk = _config.config("biocypher").get("log_to_disk", True)
+
+        if log_to_disk:
+            logdir = (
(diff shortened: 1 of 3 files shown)


biocypher/_logger.py
```diff
@@ -60,35 +60,43 @@ def get_logger(name: str = "biocypher") -> logging.Logger:
...
-        logger.addHandler(file_handler)
         logger.addHandler(stdout_handler)
 
         # startup message
         logger.info(f"This is BioCypher v{__version__}.")
-        logger.info(f"Logging into `{logfile}`.")
+        if log_to_disk:
+            logger.info(f"Logging into `{logfile}`.")
+        else:
+            logger.info("Logging into stdout.")
```
slobentanzer, line 99: @nilskre this means that logging into stdout will always only be on the info level and above, never debug, right?

should we inform users that they need to log to disk if they want debug / is that even a good idea? what about passing the log to upstream packages (ie something imports biocypher and then wants access to the debug logs)?

nilskre: Yes (if I understand the code correctly, this is equal to the behavior before in [_logger.py](https://github.com/biocypher/biocypher/blob/main/biocypher/_logger.py#L72), where the stdout_handler is set to INFO and the file_handler is set conditionally.
I agree, that the log level should be the same no matter if it is logged to disk or to stdout.

slobentanzer: The main question is how to set a debug level log for the file (or some upstream package that may want to collect all debug logs) without spamming the console with detailed debug logs. It should definitely stay the default solution to only print info and above to the console.

slobentanzer: With the newest commit, we can't have a debug log to file but higher-level console output. I think that is not practical. However, I am not sure I want two log level settings in the config...

nilskre: Ok, I have misunderstood it and consequently removed the last commit.

slobentanzer: So the current behaviour is to only print INFO and above to the console, except if the user sets `log_to_disk: true` in the config? So it is deactivated by default?

What happens if you set `log-directory` to some value, but not `log_to_disk: true`?

slobentanzer: Hm, how is the default setting 'true' implemented? The line in the standard config is commented out. If we want it to be true by default, I would just set it there, to be overwritten by the user in case they don't want it.

slobentanzer: So if we uncomment the standard in the config, we would not need this default=True setting. I think that would be more transparent. I'll push some changes.

slobentanzer (approved):
Thanks!

Issue #284, opened 2023-10-30
Bug: Running BioCypher without defined schema fails


nilskre:
Running BioCypher without a defined schema fails because the [`OntologyMapping` is set to `None`](https://github.com/biocypher/biocypher/blob/main/biocypher/_core.py#L184), which leads to `AttributeError: 'NoneType' object has no attribute 'extended_schema', caused when adding the synonym information [here](https://github.com/biocypher/biocypher/blob/main/biocypher/_ontology.py#L594).

Proposed solution: 
If the schema_config_file is not passed to the `OntologyMapping`, set the schema to an empty dict (`schema_config = {}`).

PR #286, opened 2023-10-30
Bugfixes #284 #285

nilskre:
- #284 
- #285

### DIFF
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -62,28 +62,6 @@ def test_ontology_without_schema_config(core_no_schema):
     assert isinstance(core_no_schema._ontology._nx_graph, nx.DiGraph)
 
 
-def test_show_full_ontology_without_schema_config(core_no_schema):
-    assert core_no_schema
-
-    core_no_schema._head_ontology = {
-        "url": "http://semanticweb.cs.vu.nl/2009/11/sem/",
-        "root_node": "Core",
-        "format": "rdf",
-    }
-    core_no_schema._ontology_mapping = None
-
-    core_no_schema._get_ontology()
-
-    treevis = core_no_schema.show_ontology_structure(full=True)
-
-    assert treevis is not None
-    assert "core" in treevis
-    assert "actor" in treevis
-    assert "event" in treevis
-    assert "place" in treevis
-    assert "time" in treevis
-
-
 @pytest.mark.parametrize("length", [4], scope="function")
 def test_write_schema_info_as_node(core, _get_nodes):
     core.write_nodes(_get_nodes)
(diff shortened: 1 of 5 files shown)


slobentanzer:
I did some work on enabling no-schema running around this commit: cc3af8855f1696db6d873c70acd5cb003d84cf59

Is this related? I didn't have problems running without schema, and the tests also pass. Can you give an example of failure?

nilskre:
After looking into it in depth I have found the differences between the tested case and the described problem in #284.
The test initializes BioCypher without a schema. This is possible without any problem.

The problem occurs when initializing BioCypher without a schema and then visualizing the full ontology with `show_ontology_structure(full=True)`. This leads to the problem described in the [issue](https://github.com/biocypher/biocypher/issues/284).
The tests pass because this case is currently not tested.

Does this clarify the open questions?

slobentanzer:
@nilskre Yes, clear now. However, could you please add a test (or more, as required) that allows replicating the desired behaviours? I think all changes we introduce should ideally be test-driven.

test/test_integration.py
```diff
@@ -62,6 +62,28 @@ def test_ontology_without_schema_config(core_no_schema):
     assert isinstance(core_no_schema._ontology._nx_graph, nx.DiGraph)
 
 
+def test_show_full_ontology_without_schema_config(core_no_schema):
```
slobentanzer, line 65: @nilskre I see two issues with doing it like this:
- this test repeats a lot of code of the above test, including two identical calls to an online resource (which should be kept to an absolute minimum)
- testing the ontology structure does not need to be an integration test

To fix the first, one could create a fixture that is used by both tests, but I would not do that and rather go for implementing a unit test based on one of the on-disk ontology files that are used in the ontology test module. All you need is an `Ontology` class instance without a schema definition, you can call the `show_ontology_structure` method of that class directly.

nilskre:
Yes, totally agree! Now a test for this case is added. Sorry for not directly providing a test for the changes. Will take this into account in the future.

test/test_integration.py
nilskre: Yes, true. Thanks for the feedback. I will update it (probably tomorrow).

nilskre: The test is now adapted accordingly. 
When reviewing don't be confused, that there is still a BioCypher object initialized (with a local ontology). This is necessary, because the line, which needs to be tested is in [_core.py](https://github.com/biocypher/biocypher/blob/d3195ea33d2e7b2bd516336dc1bbf38999a8221e/biocypher/_core.py#L184). I have tested it locally with the old code (new test fails) and the new code (new test passes).

slobentanzer (approved):
Thanks!

Issue #228, opened 2023-08-09
Cellosaurus


slobentanzer:
https://github.com/IGVF-DACC/igvf-catalog/blob/main/data/adapters/cellosaurus_ontology_adapter.py

PR #289, opened 2023-11-08
fix issue #228: fix doctest and run docs build on PRs

nilskre:
The doctest is fixed and to avoid failures like this in the future, the CICD pipeline is adapted to also run the docs build on PRs.


slobentanzer (approved):
Looks good, thanks!

PR #301, opened 2023-12-08
Bugfix resource caching

### LINKED ISSUE #300: Cached resource version is never used
There are two possible cases for caching:
1) use the `Downloader` without a specified cache dir (this case is tested in the corresponding unit test)
2) use the `Downloader` with a specified cache dir (this case is not tested in the corresponding unit test)

When running the unit test with a `Downloader` with a specified cache dir, they fail.

Reasons: 
- The corresponding else branch, if the cached resource should be used, is missing [https://github.com/nilskre/biocypher/blob/main/biocypher/_get.py#L124](https://github.com/nilskre/biocypher/blob/main/biocypher/_get.py#L124). Therefore the cached resource is never used.
- In the test case 1) every time a new tmp cache dir is created. Thus the reading process of a cached resource is never executed, nor tested.

nilskre:
Fix #300

### DIFF
--- a/test/test_get.py
+++ b/test/test_get.py
@@ -10,8 +10,18 @@ from biocypher._get import Resource, Downloader
 
 
 @pytest.fixture
-def downloader():
-    return Downloader(cache_dir="./.cache")  # TODO: test both cases None)
+def downloader(request):
+    return request.getfixturevalue(request.param)
+
+
+@pytest.fixture
+def downloader_without_specified_cache_dir():
+    return Downloader()
+
+
+@pytest.fixture
+def downloader_with_specified_cache_dir():
+    return Downloader(cache_dir="./.cache")
 
 
 @given(
@@ -29,11 +39,27 @@ def test_resource(resource):
     assert isinstance(resource.lifetime, int)
 
 
+@pytest.mark.parametrize(
+    "downloader",
+    [
+        "downloader_without_specified_cache_dir",
+        "downloader_with_specified_cache_dir",
+    ],
+    indirect=True,
+)
 def test_downloader(downloader):
     assert isinstance(downloader.cache_dir, str)
     assert isinstance(downloader.cache_file, str)
 
 
+@pytest.mark.parametrize(
+    "downloader",
+    [
+        "downloader_without_specified_cache_dir",
+        "downloader_with_specified_cache_dir",
+    ],
+    indirect=True,
+)
 def test_download_file(downloader):
     resource = Resource(
         "test_resource",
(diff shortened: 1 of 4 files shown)


slobentanzer:
Thanks a lot; there are some issues that prevent a merge, and due to the hackathon, I did a manual fix of the initial problem on `main` (the conversion between datetime and str was implemented inconsistently). We can get back to the PR after the hackathon.

test/test_get.py
```diff
@@ -10,8 +10,18 @@
...
+
+
+@pytest.fixture
+def downloader_without_specified_cache_dir():
+    return Downloader()
+
+
+@pytest.fixture
+def downloader_with_specified_cache_dir():
+    return Downloader(cache_dir="./.cache")
```
slobentanzer, line 24: tests should not write to the project directory. for keeping a folder around for multiple tests, we usually use pytest fixture scopes (module or session scope). can get tricky though. :)

test/test_get.py
```diff
@@ -46,20 +72,35 @@ def test_download_file(downloader):
...
     paths = downloader.download(resource)
-    # should not download again
-    assert paths[0] is None
-
-    # manipulate cache dict to test expiration (datetime format)
-    downloader.cache_dict["test_resource"][
-        "date_downloaded"
-    ] = datetime.now() - timedelta(days=8)
+    # should not download again (assert that cache folder is returned)
+    if downloader == "downloader_without_specified_cache_dir":
```
slobentanzer, line 76: these conditionals both fail, because downloader is an instance of Downloader, not a string

test/test_get.py
```diff
@@ -46,20 +72,35 @@ def test_download_file(downloader):
...
+        datetime.now() - timedelta(days=8)
+    )
 
     paths = downloader.download(resource)
     # should download again
     assert len(paths) == 1
-    assert paths[0] is not None
-
-
+    if downloader == "downloader_without_specified_cache_dir":
```
slobentanzer, line 89: similar to above, both conditionals fail here, never asserting anything

test/test_get.py
nilskre: True. :see_no_evil: Learning: never trust what ChatGPT is generating

nilskre:
Updated this PR based on the feedback and now ready for second review.

nilskre:
The [test_download_directory_and_caching](https://github.com/biocypher/biocypher/blob/main/test/test_get.py#L127) and [test_download_zip_and_expiration](https://github.com/biocypher/biocypher/blob/main/test/test_get.py#L150) are not using the `Downloader` fixture. Is this intended?

Issue #307, opened 2024-01-30
OWL/TTL ontology containing child nodes with multiple parent nodes defined via `owl:intersectionOf` are not working


nilskre:
I stumbled across this issue when working with the Snomed CT ontology (in OWL) and searching for specific terms (e.g. `Trastuzumab`). They are not present in the ontology parsed by BioCypher. 

Debugging Biocypher's ontology parsing process:
- In the rdf graph there are around 390000 concepts. 
- The parsed networkx graph only contains 67133 nodes/concepts.

Why are so many concepts missing in the parsed ontology?
-> Whenever a concept has multiple parent nodes it is not added to the networkx graph.
Example: The following example OWL class is not in the parsed ontology (and has multiple parent nodes defined in owl:intersectionOf):

```
    <owl:Class rdf:about="http://snomed.info/id/771358001">
        <rdfs:subClassOf>
            <owl:Class>
                <owl:intersectionOf rdf:parseType="Collection">
                    <rdf:Description rdf:about="http://snomed.info/id/120575009"/>
                    <rdf:Description rdf:about="http://snomed.info/id/38033009"/>
                </owl:intersectionOf>
            </owl:Class>
        </rdfs:subClassOf>
        <rdfs:label xml:lang="en">Amputation stump of lower limb (body structure)</rdfs:label>
        <skos:prefLabel xml:lang="en">Amputation stump of lower limb</skos:prefLabel>
    </owl:Class>
```

In ttl it would be defined like this:
```
###  http://snomed.info/id/771358001
<http://snomed.info/id/771358001> rdf:type owl:Class ;
                                  rdfs:subClassOf [ owl:intersectionOf ( <http://snomed.info/id/120575009>
                                                                         <http://snomed.info/id/38033009>
                                                                       ) ;
                                                    rdf:type owl:Class
                                                  ] ;
                                  rdfs:label "Amputation stump of lower limb (body structure)"@en ;
                                  skos:prefLabel "Amputation stump of lower limb"@en .
```

To understand the issue I have created a toy example.

<details>
  <summary>ontology.ttl -> working as expected</summary>

````
@prefix dc: <http://purl.org/dc/elements/1.1/> .
@prefix grddl: <http://www.w3.org/2003/g/data-view#> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
[...]

PR #308, opened 2024-01-31
Bugfix ontology multiple parents

nilskre:
Bugfix for #307 

Open to be discussed is the visualization with treelib. We now allow a node to have multiple parents. This violates the definition of a tree. Consequently, in the visualization with treelib the nodes with multiple parents are added once to the parsed ontology. All other occurrences are neglected and not added to the parsed ontology.

Two options:
1) Keep it like this and document this behavior.
2) Fix it by replacing treelib with some other library (do you know one?) or custom implementation (could get hacky). 

@slobentanzer Do you have a preference which option to choose?


slobentanzer:
Looks good, thanks. What about the third option, expanding the tree for visualisation? Since we conceptually now allow a child in multiple places, we can print all the places where it is attached as well. Or would this require a lot of memory for the expanded tree?

slobentanzer:
IIRC, when printing the regular way (only the labels that are in the config), we do a depth-first search from the child to the root to find all single paths that are part of the actual hierarchy of the KG. If there are multiple ways to get from one child to the root, we could just find all of these instead of just one, and build the visualisation from that?

https://github.com/biocypher/biocypher/blob/0388dd48ce312d934c14542620d322207ef989bc/biocypher/_ontology.py#L581

nilskre:
> What about the third option, expanding the tree for visualisation?

The tree for the visualization is created in [create_tree_visualisation](https://github.com/biocypher/biocypher/blob/main/biocypher/_misc.py#L79). The children nodes are added [here](https://github.com/biocypher/biocypher/blob/main/biocypher/_misc.py#L122). If the children node has multiple parents the `parent` variable contains a list of nodes. A loop, which duplicates them and adds all of them into the tree fails with `treelib.exceptions.DuplicatedNodeIdError: Can't create node with ID 'child'` (this is expected, due to the definition of tree and graph). When duplicating the node and creating different IDs for all nodes the visualization gets stuck into a endless loop, because it is not unambiguous any more under which duplicate new nodes should be placed.

slobentanzer:
Can we add a suffix to the child nodes that are duplicated, such that the IDs are different?

nilskre:
After some practical experiments and conceptual considerations I am in favor of choosing option 1.
Thus checking if there are multiple parents in the parsed ontology graph. If yes log a warning informing the user that some nodes have multiple parents and that this is not visualized in the plotted hierarchy tree. Furthermore one could mention that the full parsed ontology can be viewed by writing a graphml file.

Why?
Hierarchical trees are not meant to visualize arbitrarily complex graph structures. I have practically experienced this when trying to implement it (endless loops, ...).

Consider the following example:
image

- The green node `C1` has multiple parents. -> We place it under all it's parents in the ontology tree (here `B2` and `A2`). 
- For all nodes under `C1` (all orange nodes) the question arises where to place them. Following the pattern to place them at all places, where they occur, they need to be duplicated.
- With increasing graph size this can get arbitrarily complex.

Based on this I think it is the cleanest solution to add each concept only once to the hierarchical tree for the visualization.

slobentanzer:
Agree, it is probably fine for now like that, and including the warning will at least help in finding the cause for the users.

nilskre:
> @nilskre did you know that you can rerun actions manually in the actions tab? no need to do a push to trigger in those cases.

Yes, but only if I have push access to the repo (which is now the case - thanks).

nilskre:
Now with green pipeline ready for review @slobentanzer

slobentanzer (commented):
Looks great, thanks! Before we merge, would it be a good time for you to go over the ontology documentation (https://github.com/biocypher/biocypher/blob/main/docs/tutorial-ontology.md) one time, to add brief information about this recent addition, as well as generally check whether something can be improved from your point of view?

nilskre:
Great!
Yes, that is a good idea. I will go over it and update it accordingly.

nilskre:
The docs are now also updated.
@slobentanzer Are the changes to the docs understandable and clearly formulated?

slobentanzer (approved):
Great work, thanks!

PR #309, opened 2024-02-01
Brackets in ontology names

### LINKED ISSUE #306: Brackets in ontology names result in improper Neo4j file names
When names in the underlying ontology contain brackets, the generated files for the Neo4j import are not proper.

Example:
- Name in the Snomed CT ontology: `Patient (person)`
- When writing the files for the Neo4j import, BioCypher transforms this name to CamelCase (results in: `Patient (person)`).
- The generated import statement `bin/neo4j-admin import --database=neo4j --delimiter="\t" --array-delimiter="|" --quote="'" --force=true --skip-bad-relationships=true --skip-duplicate-nodes=true --nodes="/biocypher-out/20240115145536/Patient (person)-header.csv,/biocypher-out/20240115145536/Patient (person)-part.*"` fails with `java.lang.IllegalArgumentException`. [Neo4j can not handle whitespaces in filenames](https://neo4j.com/developer/kb/space-in-import-filename-for-load-csv/).

After removing the whitespace and the brackets from the file names and the import statement everything works as expected.

Is there any advice on how to handle this?
I would suggest to check for not supported characters (such as brackets) and remove them before writing the files (probably in  [name_sentence_to_pascal](https://github.com/biocypher/biocypher/blob/main/biocypher/_translate.py#L470) or somewhere in [write](https://github.com/biocypher/biocypher/blob/main/biocypher/_write.py) .

nilskre:
Fixes #306 

Added a function for checking if a string is compliant with the Neo4j naming rules. When writing the node and edge files it is checked, whether the file name is compliant with the Neo4j naming rules. If not the file name is adapted accordingly by removing the characters in question.

### DIFF
--- a/biocypher/_logger.py
+++ b/biocypher/_logger.py
@@ -48,7 +48,7 @@ def get_logger(name: str = "biocypher") -> logging.Logger:
         # create logger
         logger = logging.getLogger(name)
         logger.setLevel(logging.DEBUG)
-        logger.propagate = False
+        logger.propagate = True
 
         # formatting
         file_formatter = logging.Formatter(
--- a/biocypher/_write.py
+++ b/biocypher/_write.py
@@ -13,6 +13,7 @@ BioCypher 'offline' module. Handles the writing of node and edge representations
 suitable for import into a DBMS.
 """
 
+import re
 import glob
 
 from ._logger import logger
@@ -22,7 +23,6 @@ logger.debug(f"Loading module {__name__}.")
 from abc import ABC, abstractmethod
 from types import GeneratorType
 from typing import TYPE_CHECKING, Union, Optional
-from datetime import datetime
 from collections import OrderedDict, defaultdict
 import os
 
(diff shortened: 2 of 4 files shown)


nilskre:
> Should we replace dots with underscores categorically?

From my point of view, there is nothing that speaks against it. 
To get it right: Would this mean to 
1) adapt the source code for the implicit inheritance to already produce compliant names e.g. `KEGG_pathway`
2) or would this mean to replace the dot with an underscore in the `parse_label` method?

slobentanzer:
Good question. I think since this is also a Neo4j-specific problem, for now the less deleterious way would be to change it at the latest possible moment, which is the parsing.

Something to keep in mind: if we do that, we need to update the docs again to explain and replace the dots in the examples.

nilskre:
Sounds good. I would propose to create an issue for this and implement it in a separate PR (then code and docs can be changed at the same time). Otherwise we first need to get the changes from the other branch.

slobentanzer:
OK, I'll merge this one after the current merge has completed. You can go ahead and do the separate issue / PR.

slobentanzer (approved):

Issue #144, opened 2023-03-14
Biolink - disjoint top level classes: how to integrate?


slobentanzer:
Those become disjoint root nodes of subgraphs when I don't connect them to 'entity' manually:

'frequency qualifier mixin',
'chemical entity to entity association mixin',
'ontology class',
'relationship quantifier',
'physical essence or occurrent',
'gene or gene product',
'subject of investigation',

slobentanzer:
Could add a general function to look for the topmost entities and connect them to a virtual one if so desired. We have the same problem in the BRO.

PR #311, opened 2024-02-05
#310: improve runtime of rdflib to networkx transformation

### LINKED ISSUE #310: Ontology: Improve performance of rdflib -> networkx graph
For huge ontologies the transformation from the rdflib  graph to the networkx graph is quite slow.

nilskre:
This PR fixes #310 and #316 

Speed up the transformation from rdflib graph to networkx graph in `Ontology`

For the Biolink ontology:
- around 200 nodes in the ontology
- Old implementation: 2.23 seconds
- New implementation: 0.04 seconds

For Snomed CT:
- over 200000 nodes in the ontology
- Old implementation: didn't converge on my computer
- New implementation: 47.79 seconds

The improved solution currently differs in the handling of mixins.
As far as I understand this the mixins describe properties, that should be present at specific ontology terms (e.g. if an ontology class is a subclass of the `gene expression mixin` it is expected, that the instances of this ontology term have this property.

In the setting of the Biolink ontology with root node `entity` the former and this solution differ. The former solution has all ontology terms under `entity` and all ontology terms under `Mixin`. This solution only has all terms under `entity` (the terms under `Mixin` are not included, because they are not under entity).

Are the mixins further used by BioCypher later on? And is there any advice from your side how to handle this? Thanks in advance! @slobentanzer 

image


nilskre:
I just realized, that you (@slobentanzer) already had the same underlying question #132 (with the related issue #144 )

Maybe now is a good moment to discuss this.

### What are mixins  
Mixins refer to a concept or property that can be reused across multiple ontologies or ontology modules (e.g. specific property present on a concept). Mixins are not instantiated directly. To use them the concept of the ontology is a subclass of this mixin.

### The BioCypher use case
BioCypher uses ontologies to ground the knowledge in the underlying domain. In the `Translator` BioCypher checks, if the data, which should be inserted has a correspondence in the ontology. 
The structure of the data (which is usually defined by mixins) is not checked at the moment (in theory this is also possible, but this is a huge effort and I don't see the need for this at the moment).

### Conclusion
Due to the fact, that one would not directly instantiate a mixin it should be safe to remove the mixins from the BioCypher ontology representation.
Consequently in my opinion mixins can be removed from the internal BioCypher ontology representation.

If this alright, I will adapt this PR accordingly.

slobentanzer:
Agree it's good to discuss now, although I will need to focus on BioChatter until the paper is out, so I can get into this maybe next week, at the earliest.

I am currently focusing on running and evaluating the benchmark, do you have time to extend the use cases?

slobentanzer (approved):
@nilskre Good job, although there is a question why there are so many `none` type nodes are produced from the RDF files. I have for now added a conditional to get rid of them and will merge here. Will create a new issue for this question (#318).

PR #321, opened 2024-03-05
Neo4j version 5 patch

mbaric758:
Adapt Neo4j syntax and functions according to detected version.


slobentanzer:
Hi @mbaric758,
many thanks, will review ASAP. :)

mbaric758:
> Hi @mbaric758 , thanks for this PR! Just to clarify the current state of this PR: Is this at the moment work in progress (due to the failing tests) or already in a state to be reviewed? Thanks in advance!

Hi @nilskre ,

I also did a pull request on neo4j-utils concerning the v5 patch ( https://github.com/saezlab/neo4j-utils/pull/5 ), so we should probably wait for that to be resolved first. 

Thanks!

nilskre:
Great, thanks! Then we continue here, as soon as the other PR is resolved.

It would be great if the [CICD-Pipeline on PRs](https://github.com/biocypher/biocypher/blob/main/.github/workflows/pr.yaml) could be extended to be executed once with neo4j v4 and once with neo4j v5. Then upward and downward compatibility is automatically ensured.
To do this:
- Therefore you can **add the neo4j version to the test matrix** [here](https://github.com/biocypher/biocypher/blob/main/.github/workflows/pr.yaml#L11). This then automatically creates the test matrix seen here on this PR. 
- **Pass the defined neo4j version to the test action** [here](https://github.com/biocypher/biocypher/blob/main/.github/workflows/pr.yaml#L41). An example how to pass variables to a github action can be seen [a few lines above](https://github.com/biocypher/biocypher/blob/main/.github/workflows/pr.yaml#L35). And [here](https://github.com/biocypher/biocypher/blob/main/.github/actions/install/action.yaml#L3) you can see how to use this variable in a github action.
- Finallly adapt the [**docker command** to start neo4j](https://github.com/biocypher/biocypher/blob/main/.github/actions/test/action.yaml#L15) to start neo4j with the version defined in the variable. 

If you have any further questions don't hesitate to ask. Furthermore, if you need support doing this just hit me up. (:

mbaric758:
Hey @nilskre , the [PR on neo4j-utils ](https://github.com/saezlab/neo4j-utils/pull/5) is resolved so I updated the testing pipeline as you suggested (it should be double-checked if that's correct). The pytest tests should be ok for both versions of neo4j now.

nilskre:
Hey @mbaric758 , Thanks a lot! I will review it asap.

nilskre:
Just for information/documentation the current state on this: The CICD-Pipeline is now up and running again.
In case, that neo4j is running locally on the same machine, everything should work as expected.
In case, that neo4j is running in a Docker container (like in the testing setup), the process to get the neo4j version fails (see [run](https://github.com/biocypher/biocypher/actions/runs/8361119749/job/22888431399?pr=321): tries on the CICD runner to get the neo4j version -> fails; should do this within the docker container).
Will continue soon on this to adapt it somehow to maybe get the neo4j version from the API.

nilskre:
To inform about the current state: This will be ready for merging soon. It just waits fo [this PR in neo4j-utils](https://github.com/saezlab/neo4j-utils/pull/8) to be resolved, and then this will be merged.

JPonsa:
@mbaric758 not sure if this would hanging fruit.

It would be great if the neo4j syntax could be changed based on operative system.

To adapt the neo4j-admin-import-call for windows you have to change bin\neo4j-admin for bin/neo4j-admin and remove the setting --quote='"'

mbaric758:
Hi @JPonsa , thanks for the suggestion! I think it should be relatively straightforward, but would probably require changing the tests and neo4j-driver modules again. Would you suggest implementing that here, or doing another PR for that, since we seem to be having a version that works now?

JPonsa:
Thanks @mbaric758 . I don't know if it is better to have it as a different PR or in the same PR.   A question for the biocypher team. @nilskre

nilskre:
Hello,
thanks @JPonsa for your suggestion. And thanks @mbaric758 for the time estimation.
The behavior you described is not directly related to this Neo4j v5 version support PR. Therefore, I would suggest to create a separate issue/PR for this. Issue/PR are welcome.

JPonsa:
@mbaric758 

The comment below is an issue with neo4j not biocypher. Mentioning so (i) people is aware (ii) see if we are able to find a workaround 

I discovered that somehow, at least on Windows, there is a max length for the neo4j-admin database import command. I am assuming this applies to all commands from the neo4j terminal. In my case importing a large list of files and having a full path for each file meant that I exceeded this limit. The command 2nd half of the command was getting stripped and failing to execute. I had to move the data into the dbms folder (the same folder you have the neo4j bin\) so I could use shorter relative paths. I tried splitting the command in 2 (e.g. nodes and relationships) but failed. I had the issue even trying to execute the command as script.  Doing a quick google search seems this limit could be 8,191 characters, although I found references to other values.

It would be nice to have at least an alert/message somewhere fighting if the admin import command produced exceeded this limit

nilskre:
Thanks for the comment @JPonsa 
Your comment is not related to this PR. For a further discussion please open a new issue.

biocypher/write/graph/_neo4j.py
```diff
@@ -22,13 +24,45 @@ class _Neo4jBatchWriter(_BatchWriter):
...
+        super().__init__(*args, **kwargs)
+
+        try:
+            neo4j_version = _n4jversion.Neo4jVersion()
+            neo4j_version = neo4j_version.version
+        except:
+            logger.info(
+                "Not able to determine Neo4j version. Use default (version 4)."
+            )
+            neo4j_version = 4
```
nilskre, line 47: Great work so far @mbaric758 !
After carefully reviewing both the online and offline cases, I noticed one aspect that I think we should address together:
- When it comes to the online mode, everything seems to be running smoothly. The neo4j database version is directly fetched from the driver, and the correct import call is utilized. So, no issues there.
- However, in the offline mode, the version determination becomes a bit tricky. At the time of creating the Knowledge Graph (KG), we might not know whether the user intends to insert the KG into Neo4j v4 or v5. To make things more user-friendly, it would be advantageous if the generated import call could determine the version at runtime and use the appropriate command.

To tackle this, I could imagine the following:
- We could integrate an if-else condition within the `neo4j-admin-import.sh` script, allowing it to dynamically choose the appropriate command based on the detected version.
- Alternatively, we could create two separate scripts: `neo4j-admin-import.sh` and `neo4j-v5-admin-import.sh`, each tailored to handle the import process for the respective versions.

What do you think about this @mbaric758 ?

mbaric758: Hi @nilskre , thanks for the review! 

Probably the most user friendly way of doing it would be to go with your first suggestion - implementing an if-else condition within `neo4j-admin-import.sh` , so the whole process of import call generation is automatic based on Neo4j version.

nilskre: The issue with neo4j-utils is now fixed (by using the latest working version `0.0.7` and implementing the rest directly in BioCypher).
Furthermore, it is now implemented that both import calls for v4 and v5 are written into the `neo4j-admin-import.sh`. At the moment there is not yet a if else in the import script.

So we are on the home straight.
The last open points are:
- Implement the if else in the script?
- Adapt the tests in `test/write/graph/test_neo4j.py` to test if the import script is correct for v4 and v5.
- Update the [neo4j docs](https://biocypher.org/output/neo4j.html) according to the changes.

@mbaric758 Can you verify my changes and have a look at open points?

Thanks a lot in advance!

mbaric758: Thanks @nilskre ! I implemented an if-else statement for the import script which checks Neo4j version and then uses the appropriate import call. It doesn't seem to change the way the tests behave in `test/write/graph/test_neo4j.py` . Have a look and tell me what you think. Thanks!

biocypher/write/graph/_neo4j.py
```diff
@@ -285,7 +285,11 @@ def _construct_import_call(self) -> str:
         import_call_neo4j_v5 = self._get_import_call(
             "database import full", "", "--overwrite-destination="
         )
-        import_script = f"# Neo4j v4 import call:\n{import_call_neo4j_v4}\n# Neo4j v5 import call:\n{import_call_neo4j_v5}\n"
+        neo4j_verison_check = (
+            f"version=$(neo4j-admin --version | cut -d '.' -f 1)"
```
nilskre, line 289: On my system I get as output for `version=$(neo4j-admin --version | cut -d '.' -f 1)`
`WARNING: Max 1024 open files allowed, minimum of 40000 recommended 4`. This will lead to failure in the next line. So it would be great to adapt the bash command to be more specific with the version selection.

biocypher/write/graph/_neo4j.py
```diff
@@ -285,7 +285,11 @@ def _construct_import_call(self) -> str:
...
+        )
+        output_command = '"${cmd[@]}"'
+        import_script = f"#!/bin/bash\n# Check Neo4j version\n{neo4j_verison_check}\nif [ $version -ge 5 ]; then\n\t # Neo4j v5 import call:\n\t{import_call_neo4j_v5}\nelse\n\t# Neo4j v4 import call:\n{import_call_neo4j_v4}\nfi\n{output_command}"
```
nilskre, line 292: Isn't there a `\t` missing before `{import_call_neo4j_v4}`

nilskre, line 292: Have you tested the whole import script?
I just tested it with one of the small tutorial examples within BioCypher. After copying the script to the DB root and running it I get `./import.sh: line 4: [: too many arguments`.
The import script looks like this:
```
#!/bin/bash
# Check Neo4j version
version=$(neo4j-admin --version | cut -d '.' -f 1)
if [ $version -ge 5 ]; then
	 # Neo4j v5 import call:
	bin/neo4j-admin database import full --delimiter=";" --array-delimiter="|" --quote="'" --overwrite-destination=true --nodes="pathes" relationships 
else
	# Neo4j v4 import call:
bin/neo4j-admin import --delimiter=";" --array-delimiter="|" --quote="'" --force=true --nodes="pathes" --nodes=" --database=relationships 
fi
"${cmd[@]}"
```

biocypher/write/graph/_neo4j.py
```diff
@@ -285,7 +285,11 @@ def _construct_import_call(self) -> str:
         import_call_neo4j_v5 = self._get_import_call(
             "database import full", "", "--overwrite-destination="
         )
-        import_script = f"# Neo4j v4 import call:\n{import_call_neo4j_v4}\n# Neo4j v5 import call:\n{import_call_neo4j_v5}\n"
+        neo4j_verison_check = (
+            f"version=$(neo4j-admin --version | cut -d '.' -f 1)"
+        )
+        output_command = '"${cmd[@]}"'
```
nilskre, line 291: What is this doing?

nilskre (commented):
Thanks for adapting it! There are some questions left on my side. Can you clarify it? Thanks in advance!

biocypher/write/graph/_neo4j.py
nilskre: Regarding the tests: E.g. [here](https://github.com/biocypher/biocypher/blob/main/test/write/graph/test_neo4j.py#L56) and [here](https://github.com/biocypher/biocypher/blob/main/test/write/graph/test_neo4j.py#L110) we assert, how the import script should look like. In [this commit](https://github.com/biocypher/biocypher/pull/321/commits/c85e52f11c667195476b68753032cbe1e0f256b8) you have adapted this part of the tests to pass for the new adapted import call. It would be great, if you can adapt this to check if the import call looks like expected for v4 and v5 (all in one script). At the moment it only checks if some general words are in the import call.

biocypher/write/graph/_neo4j.py
mbaric758: This may be related to your settings for the number of open files allowed? The recommended number by Neo4j is [40000 at least](https://neo4j.com/developer/kb/number-of-open-files-on-linux/
).

biocypher/write/graph/_neo4j.py
mbaric758: It was left by mistake, sorry.

biocypher/write/graph/_neo4j.py
nilskre: Nice, thanks for the link. Now I have fixed my local config.

biocypher/write/graph/_neo4j.py
mbaric758: I added a test to `test/write/graph/test_neo4j.py` that checks whether all the distinct elements of both import calls are present within the generated import call, so have a look and tell me what you think.

nilskre: Looks good, thanks!

nilskre (approved):

nilskre:
Great work! Thanks a lot @mbaric758 and @jdreo

PR #328, opened 2024-03-13
doc(error messages): more detailled errors

jdreo:
- More hints as to what could be wrong in Ontology.
- Replaces reference to Biolink with "ontology".


nilskre (approved):

nilskre:
Good work, thanks!

PR #329, opened 2024-03-13
refactor(write): return the path of the import call file

jdreo:
This allows the user to manipulate the actual path, since no accessors exist otherwise. For instance, this can be printed on stdout and piped in a shell automatically. The returned string can still be used in a test, hence not changing the interface much.


nilskre (approved):

nilskre:
Good work, thanks!

PR #338, opened 2024-04-18
Rdf output adapter

slobentanzer:
This PR introduces an RDF output module based on `rdflib`


slobentanzer:
@LoesvdBiggelaar @nilskre here is the new PR

nilskre (commented):
Thanks for addressing the feedback. @LoesvdBiggelaar 
There is only one comment regarding the test left.

And one other question: How do you further process the RDF output? Do you load it into some kind of triple store / visualize it somewhere?  
The context of my question is, that for the other output adapters the created Knowledge Graph can be loaded and visualized somewhere (e.g. in Neo4j or PostgreSQL). This is directly supported by BioCypher through the generated import script. I asked myself if this would somehow also be possible for RDF?

nilskre (dismissed):
Great work, thanks! @LoesvdBiggelaar 
@slobentanzer From my point of view this is ready to be merged.

nilskre:
> For now you can use the ttl/xml (or other format) to load it into any application you want to use. I tested the RDFWriter by running it with the Open Targets adapter and loading the output files into GraphDB. In GraphDB I checked if I'm satisfied with my graph. I manually added other ontologies to expand my KG, which worked perfectly.
> loading it in GraphDB is indeed a manual process, for now at least. I quickly looked around and saw that there is e.g. a [docker image of graphDB](https://hub.docker.com/r/ontotext/graphdb/), which could be used. I didn't check it out in detail but it looks like you can then create a pipeline similar to that of Neo4j.

Thanks for the helpful explanation.

> I think this would indeed be a valuable next step. maybe a ticket can be created for this?

Definitely, feel free to open an issue for this.

nilskre:
To get this merged, I have implemented a temporary CICD-Fix.

slobentanzer:
Many thanks both, merging now. :)

PR #340, opened 2024-04-22
#339: adapt pr cicd config to limit number of runs for neo4j v4 and v5

### LINKED ISSUE #339: Run Neo4j v4 test action only on Linux (legacy)
Running both v4 and v5 tests on all OSs doubles the number of test runs on the Actions runners. I think we could be economical and treat v4 as a legacy test to run only on a subset of the matrix (Linux?).

nilskre:
Closes #339


slobentanzer (approved):
Great, thanks!

PR #360, opened 2024-07-10
Allow for time stamped directories as used in some of the examples

Mjvolk3:
The purpose of the pr is allow for csv file writing to in time stamped directories. This has the advantage of allowing for versioning of data imports. I will go through step by step.

This allows for `None` to give the default behavior of before.

```python
if import_call_file_prefix is not None:
```

In my config I have `import_call_file_prefix: "./"`.

```python
>>>print(import_call_file_prefix)
'./'
```

We then get the relative path to where the .csv files are written.

```python
>>>print(self.outdir)
'/Users/michaelvolk/Documents/projects/torchcell/biocypher-out/2024-07-10_13-11-36'
>>> print(temp_path)
'biocypher-out/2024-07-10_13-11-36'
```
We then reset the prefix to this path.

```python
>>> print(self._import_call_file_prefix)
'./biocypher-out/2024-07-10_13-11-36'
```

This allows for the `neo4j-admin-import-call.sh` to look like this... Since we don't know the date-time stamp ahead of time we cannot add it to the config so this must be done within the `_write.py`. 

```bash
/var/lib/neo4j/bin/neo4j-admin import --database=torchcell --delimiter="\t" --array-delimiter="|" --quote="'" --force=true --skip-bad-relationships=true --skip-duplicate-nodes=true --nodes="./biocypher-out/2024-07-08_18-16-00/Temperature-header.csv,./biocypher-out/2024-07-08_18-16-00/Temperature-part.*" --nodes="./biocypher-out/2024-07-08_18-16-00/Environment-header.csv,./biocypher-out/2024-07-08_18-16-00/Environment-part.*" .........
```

Otherwise we would get this.

```bash
... --nodes="./Temperature-header.csv ...
```

I am currently using this change, but any change that achieves the same outcome allowing for time stamped directories would be nice.

I have left debug statements, but they can be removed.


slobentanzer:
@nilskre I am limited in time, would you like to take a look?

nilskre:
Thanks for your PR.

I am not sure if I understand the purpose correctly. Therefore some questions and comments:

> allow for csv file writing to in time stamped directories

This is already the default case at the moment. BioCypher creates time-stamped output directories (e.g. default would be `20240710131136`).  

> This has the advantage of allowing for versioning of data imports.

Isn't this already possible with the current time-stamped output directories? Within the time-stamped directory there is the import script -> versioned by the time-stamp.

After reading the remaining description: Your change targets the handling of paths within the neo4j import script, right?
At the moment in the default case (import_call_file_prefix not set) the neo4j import script contains the full path. The import script is written inside the time-stamped output directory, next to the other node and edge files.
There are two usage scenarios:
- Run the import script from anywhere with the full paths
- Run the import script from within the time-stamped output directory with `./node.csv` and `./edge.csv`

Can you elaborate a bit more, on what your use case is?
So let's first discuss the purpose of this PR before adapting it to be mergeable.
Nevertheless already a little reminder of the [contributor/developer guide](https://github.com/biocypher/biocypher/blob/main/DEVELOPER.md). Before merging anything there need to be tests for the new features and if the user-facing behavior changes the docs need to be adapted.

Thanks in advance.

Mjvolk3:
Probably should have submitted an issue first since path changes can be pretty serious breaking changes and I still don't have a full understanding of the lib. I'll try to clarify why I'd like the change. I think we would probably need to think through additional changes that would be necessary. If you think it is desirable and we can chart out a plan I can go through adding tests, updating docs etc. 

I will reference the [pole](https://github.com/biocypher/pole) project as an example. If you run `pole/create_knowledge_graph.py` it will properly time stamp the directories in the `biocypher-out` dir each of which contains the `neo4j-admin-import-call.sh` which can allow for easy import of the database depending on the dated knowledge graph. The issue I'd like to solve is related to dates imports using the docker container.

`import_call_file_prefix: /data/build2neo` is specified in the `config/biocypher_docker_config.yaml`. This is config is responsible for getting the path right. Inside the container we can look at the import script and see that paths are correct for finding the data.

```bash
root@a3671e4553cd:/var/lib/neo4j/data/build2neo# cat neo4j-admin-import-call.sh
bin/neo4j-admin import --database=neo4j --delimiter="\t" --array-delimiter="|" --quote='+' --force=true --skip-bad-relationships=true --skip-duplicate-nodes=true --nodes="/data/build2neo/Crime-header.csv,/data/build2neo/Crime-part.*" --nodes="/data/build2neo/Officer-header.csv,/data/build2neo/Officer-part.*" --nodes="/data/build2neo/Object-header.csv,/data/build2neo/Object-part.*" --nodes="/data/build2neo/PhoneCall-header.csv,/data/build2neo/PhoneCall-part.*" --nodes="/data/build2neo/Schema_info-header.csv,/data/build2neo/Schema_info-part.*" --nodes="/data/build2neo/Person-header.csv,/data/build2neo/Person-part.*" --nodes="/data/build2neo/Location-header.csv,/data/build2neo/Location-part.*" --relationships="/data/build2neo/INVOLVED_IN-header.csv,/data/build2neo/INVOLVED_IN-part.*" --relationships="/data/build2neo/RECEIVED_CALL-header.csv,/data/build2neo/RECEIVED_CALL-part.*" --relationships="/data/build2neo/IS_RELATED_TO-header.csv,/data/build2neo/IS_RELATED_TO-part.*" --relationships="/data/build2neo/KNOWS-header.csv,/data/build2neo/KNOWS-part.*" --relationships="/data/build2neo/INVESTIGATED_BY-header.csv,/data/build2neo/INVESTIGATED_BY-part.*" --relationships="/data/build2neo/OCCURRED_AT-header.csv,/data/build2neo/OCCURRED_AT-part.*" --relationships="/data/build2neo/LIVES_AT-header.csv,/data/build2neo/LIVES_AT-part.*" --relationships="/data/build2neo/PARTY_TO-header.csv,/data/build2neo/PARTY_TO-part.*" --relationships="/data/build2neo/MADE_CALL-header.csv,/data/build2neo/MADE_CALL-part.*" 

root@a3671e4553cd:/var/lib/neo4j/data/build2neo# ls
Crime-header.csv             IS_RELATED_TO-header.csv   Location-header.csv      Object-header.csv     Person-header.csv          Schema_info-header.csv
Crime-part000.csv            IS_RELATED_TO-part000.csv  Location-part000.csv     Object-part000.csv    Person-part000.csv         Schema_info-part000.csv
INVESTIGATED_BY-header.csv   KNOWS-header.csv           MADE_CALL-header.csv     Officer-header.csv    PhoneCall-header.csv       neo4j-admin-import-call.sh
INVESTIGATED_BY-part000.csv  KNOWS-part000.csv          MADE_CALL-part000.csv    Officer-part000.csv   PhoneCall-part000.csv      schema_info.yaml
INVOLVED_IN-header.csv       LIVES_AT-header.csv        OCCURRED_AT-header.csv   PARTY_TO-header.csv   RECEIVED_CALL-header.csv
INVOLVED_IN-part000.csv      LIVES_AT-part000.csv       OCCURRED_AT-part000.csv  PARTY_TO-part000.csv  RECEIVED_CALL-part000.csv
```

Since this path is fixed the data files are always overwritten in `data/build2neo`. This is the issue I'd like to solve. As pointed out above we have all the date-versioned knowledge graphs in the `biocypher-out` dir and we could just copy these into the container, but then we would also need to make sure that we corrected `neo4j-admin-import-call.sh` as the paths will be wrong. My idea is to keep the data versioning used for both local and the docker container.

With the changes I've described `self._import_call_file_prefix = '/data/build2neo/biocypher-out/20240725174134'`. Then you could separate out builds from import and you could revert to a previous build just by running the import file under the different date-versioned dir. 

My use case is as follows. I am continuing to add new datasets to our current knowledge graph and sometimes the new data breaks things or we update the underlying schema so old queries no longer work. To get the old data out we would like to just redo bulk import of the previously built data then run the query again without having rebuild the csv files since this currently takes hours to days.  

To answer your questions. 

>Your change targets the handling of paths within the neo4j import script, right?

Yes, but it also relates to the `output_directory`. Typically I don't specify this, then I map the `biocypher-out` dir like this `-v "$(pwd)/database/biocypher-out:/var/lib/neo4j/biocypher-out" \`.

> Isn't this already possible with the current time-stamped output directories? Within the time-stamped directory there is the import script -> versioned by the time-stamp.

I pretty sure this is just for local, but if this can work with docker please let me know.

slobentanzer:
Hi @Mjvolk3,
thanks for explaining, this makes sense. There are some other problems with having a fixed path, such as having to make sure to remove previous volumes to not get an error when running the Docker build again; I think the files are not overwritten though, as you state. I found that they are just appended (creating additional parts). Agree that it would be valuable to have a generic solution for defining the Docker import paths.

However, I am not sure whether we should touch the original implementation, or rather do something with the way the Docker build works. What you describe is a valid use case, but I am not sure how flexible Docker is with variables for volume mounts. We require the `build2neo` directory not only for the paths in the import script, but also to tell docker which folder to mount / import (in the `import.sh` script). That is why we specify the folder in the `biocypher_docker_config.yaml` in the first place. Maybe the best way forward would be to develop a Docker workflow that skips the build step and rather takes an existing data folder as input for the import step (which seems to be your use case). Or to introduce a variable to stand in for the `build2neo` in the config and script (not sure how this should be implemented). Does this make sense?

Mjvolk3:
>  I think the files are not overwritten though, as you state. I found that they are just appended (creating additional parts)

Showing this to confirm what you saw. These are separate runs of the pole docker container. The only issue is that these now look like duplicate entries to neo4j, which can create some confusion.

pole csv import


>Maybe the best way forward would be to develop a Docker workflow that skips the build step and rather takes an existing data folder as input for the import step (which seems to be your use case).

This is my use case. 

> Or to introduce a variable to stand in for the build2neo in the config and script (not sure how this should be implemented). Does this make sense?

Not sure what you meant by "to stand in for." A special key word?

One other idea is to have a config for which db to import. This could be defaulted to `latest` and would just import from the latest date. If you wanted to import a previous db you would already have the exact date and could specify that in the config. `import.sh` would be able to implement this logic. But this would still require changing the current default paths which would be a headache for everyone else.

On a related note, I would like to completely decouple build from deploy to allow for better managing of resources. The reason being that the database deploy container needs far fewer resources than build. Also you can avoid using a docker container for builds in this way. This was my original use case as we were not allowed to use docker on our previous shared compute resources. Also with this I can trouble shoot multiple builds prior to deploying. I have done this in my current project. I just build to `biocypher-out/{date-time}`, create a binded mount to this dir, trouble shoot build logs etc, and when things are working I redo bulk import to the deploy container that has only a few cpu for querying. 

I am unsure how to proceed. I'd be happy to implement changes that align with your goals and support some of the use cases outlined. Thanks 😊

slobentanzer:
> Not sure what you meant by "to stand in for." A special key word?

A variable in the CLI workflow that allows specification of any folder to be used instead of hardcoding `build2neo`. I think that would be fairly tricky though, maybe we can get away with an easier solution.

> One other idea is to have a config for which db to import. This could be defaulted to latest and would just import from the latest date.

> I would like to completely decouple build from deploy to allow for better managing of resources.

This is what I meant previously; considering your use case, it would probably be most convenient to develop isolated workflows (could be Docker orchestrations) for building and deploying. We could, for instance, start with the simple case of using outputs generated by running BioCypher on host (without Docker), and then an isolated import-deploy workflow using something similar to the current Docker Compose (with the `latest` behaviour you mentioned). The only thing we would need to establish for this is a programmatic way of selecting the `latest` (or any) folder in the output directory. 

The manual way of doing this would be to remove the `build` stage of the current compose, and specifying the folder to copy into the import stage manually in the `docker-compose.yml`. You would of course have to make sure that the `import_call_file_prefix` of your `neo4j` config is correct with respect to the location you will copy your file to in the `import` container. Is it clear what I am trying to say? Do you think you can attempt that, or should we help with something? If I am not missing something, this should be possible without changing BioCypher path config.

Mjvolk3:
yes I think this makes sense. I can give it a go. I'll follow the developer guide. I would like to work on implementing this with an example. Should I do this with pole then submit PR's to pole? Or create a new `pole` something like `pole-modular`? Let me know your preference and then we can close this. Thanks.

slobentanzer:
I think you can start out modifying `pole` directly. If it works as I planned, should just be an additional Docker Compose configuration, maybe a bash script, or an additional `biocypher_config.yaml`; so it shouldn't interfere with the BioCypher part of the repo.

slobentanzer:
Closing due to implementation elsewhere; many thanks

PR #361, opened 2024-07-12
Streamline output api

### LINKED ISSUE #358: Directly get in memory networkx graph from BioCypher
Similar to the `to_df` for CSV it should be possible to get the networkx graph in-memory from BioCypher.
As part of this the interface for running BioCyher in online mode for CSV and Networkx can be streamlined
If dbms = csv | networkx and online mode, it should be possible to get the CSV and networkx representation in memory from BioCypher. For CSV this is already possible via `to_df`.

nilskre:
Closes #358 

Implement the proposed changes from [this comment](https://github.com/biocypher/biocypher/issues/358#issuecomment-2182852986) to streamline the output API.

@slobentanzer Now ready for review


slobentanzer:
@nilskre looks very good, will review asap. Does it make sense to add some developer docs for the case that someone wants to add another class of online or offline output adapter? If yes, could you draft one?

nilskre:
Great, thanks.
For now, I think the source code structure, the adapted docstrings, and the API docs are a solid starting point for the development of new output adapters.
Is this alright for you?

slobentanzer:
@nilskre fair enough; I still need to review but I can also make changes to the docs as I go if I think there need to be changes. Currently focussing on the BioChatter revision, so will be in a couple of weeks. :)

biocypher/_core.py
ryxx0811, line 1: Are the `add,` `write,` and `merge` functions intended for different modes and must they be retained? They are doing the same thing by calling the `_add_nodes` or `_add_egdes`, while the execution of these two functions is completely controlled by these two parameters.

ryxx0811 (commented):
I have one question. When the mode is offline,  `_add_nodes` and `_add_edge` call `write_nodes` and `write_edges` respectively. But when dealing with networkx, from the user's perspective, shouldn't these two methods also write networkx to file? @slobentanzer.

test/test_core.py
```diff
@@ -106,6 +111,71 @@ def test_show_full_ontology_structure_without_schema():
     assert "lethal variant" in treevis
 
 
+def test_in_memory_kg_only_in_online_mode(core):
+    for in_memory_dbms in IN_MEMORY_DBMS:
+        core._dbms = in_memory_dbms
+        core._offline = True
```
ryxx0811, line 117: Shouldn't _offline here be False?

nilskre: Indeed, for correct behavior (online mode and in memory) _offline needs to be False. Nevertheless in this test it is tested if an error is raised and the user receives a meaningful error message, if the config is not compliant (e.g. an in memory DBMS is configured in offline mode and the user tries to get an in memory instance of the KG (which is not supported; this is only supported with the config in memory DBMS in online mode).

ryxx0811: Ah, I see, thank you

biocypher/_core.py
```diff
@@ -243,20 +244,93 @@ def _get_writer(self):
...
+    def _add_nodes(
+        self, nodes, batch_size: int = int(1e6), force: bool = False
+    ):
+        """Add nodes to the BioCypher KG.
+        First uses the `_translator` to translate the nodes to `BioCypherNode` objects.
+        Depending on the configuration the translated nodes are then passed to the
+        - `_writer`: if `_offline` is set to `False`
+        - `_in_memory_kg`: if `_offline` is set to `False` and the `_dbms` is an `IN_MEMORY_DBMS`
+        - `_driver`: if `_offline` is set to `True` and the `_dbms` is not an `IN_MEMORY_DBMS`
+        """
```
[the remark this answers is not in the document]
nilskre: Yes, the driver does not exist in offline mode.
The writer can exist in offline (writes KG to file) and online mode (holds the KG in memory).

biocypher/_core.py
```diff
@@ -243,20 +244,93 @@ def _get_writer(self):
...
+        return passed
+
+    def _add_edges(self, edges, batch_size: int = int(1e6)):
+        """Add edges to the BioCypher KG.
+        First uses the `_translator` to translate the edges to `BioCypherEdge` objects.
+        Depending on the configuration the translated edges are then passed to the
+        - `_writer`: if `_offline` is set to `False`
+        - `_in_memory_kg`: if `_offline` is set to `False` and the `_dbms` is an `IN_MEMORY_DBMS`
+        - `_driver`: if `_offline` is set to `True` and the `_dbms` is not an `IN_MEMORY_DBMS`
+        """
```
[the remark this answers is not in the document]
nilskre: Same comment as above. Hope this clarifies your question (:

test/test_core.py
nilskre: No worries (:

biocypher/_core.py
nilskre: Yes completely true. This is a result of the discussion here (https://github.com/biocypher/biocypher/issues/358#issuecomment-2182881483). This is an open question to be discussed with Sebastian

biocypher/_core.py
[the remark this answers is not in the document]
slobentanzer: I think that refers to the docstring: it should be 

`_writer`: if `_offline` is set to `True` (not `False`), and 

`_driver`: if `_offline` is set to `False` (not `True`).

biocypher/_core.py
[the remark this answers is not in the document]
slobentanzer: Same as above, true and false are flipped in the docstring as is

biocypher/_core.py
slobentanzer: Yes; I think this may be a good opportunity to standardise what online and offline means, while keeping the API intuitive (e.g., write should only result in actual writing to disk). I will create a more comprehensive response below.

slobentanzer:
@nilskre, @ryxx0811, thanks for your input. I think this is a good opportunity to address the `online/offline` question systematically while keeping the user-facing API minimal and intuitive. @ecarrenolozano, this may be a "good first issue" for you to onboard into our ways of working. The problem this PR aims to solve originates in [this issue](https://github.com/biocypher/biocypher/issues/358#issuecomment-2182881483), which we opened in response to connecting with a downstream package from the lab (NetworkCommons).

Independently of the connection, it would be valuable to have a consistent implementation and user-facing API for the different ways in which we can give an output dataset (most of the time, a knowledge graph). As a basic principle, we currently have the setting of `offline`, which determines whether we write something to disk (if `True`) or not. If not, this means we are either connecting to something (initially, this was a running Neo4j DBMS), or we just stay in-memory completely (e.g., using Pandas). Current docs [here](https://biocypher.org/output/index.html). I think this can be cleaned up and refactored to be a little more intuitive, which is what we should do in this PR.

## Premises

- if we set `offline: true`, we write something to disk
- conversely, if `false`, we expect to not write anything to disk directly (although we may write something to a running DBMS somewhere)
- the user-facing API for adding nodes and edges should remain intuitive: we use `add` if we don't write, and `write` if we do. Discussion: we could also discuss whether it would not be easier to abandon different python method names and only rely on the configuration to determine whether we `add` or `write` nodes and edges. Would be great to hear your opinions on that. Would the user ever be in a situation where it would be useful to have control over whether to `add` or rather `write` something?

## Formats

Here is my understanding of the formats we currently deal with; the table indicates availability online/offline and also whether the same protocol is used to access online and offline modes. Neo4j uses different protocols (admin import and bolt driver) for offline and online.

| Format     | Online   | Offline  | Protocol    |
|------------|----------|----------|-------------|
| Neo4j      | Yes      | Yes      | different   |
| PostgreSQL | Yes      | Yes      | same        |
| ArangoDB   |   No       | Yes      |             |
| RDF        |    No      | Yes      |             |
| CSV        |     No     | Yes      |             |
| SQLite     | Yes      | Yes      | same        |
| Pandas     | Yes      |  (pickled?)        | (same?) |
| NetworkX   | Yes      |  (pickled?)        | (same?)  |

## Questions

- [] Is the above table correct? Does it reflect current capabilities? What does the ideal table look like?
- [] Should we harmonise the settings, and how?
- [] Which settings should be controlled from the config, which from the Python API (from the user perspective)?
- [] Which is the simplest and most intuitive approach for the user? 
- [] Should we simplify the Python API to only have one method for adding entities/relationships to the KG (as opposed to having `add`, `write`, and `merge` methods)?
- [] If so, do we (the users) need manual control over the write process in the offline case?

Happy for input from all of you on all of that. Many thanks :)

ryxx0811:
@slobentanzer I like the idea that the behavior is completely controlled by `offline` and `dbms`.  To keep names intuitive (`add`, `write` and `merge`), what do you think making Biocypher as an abstract class , with three subclasses  representing three different modes? There might be specific methods that need to be implemented for each mode in the future:)

slobentanzer:
@ryxx0811 why three? can you draw a diagram (UML) of the architecture?

ryxx0811:
@slobentanzer  Of course, I will do that.

ecarrenolozano:
Let me first understand the table:

**Online vs. Offline:** 

-  **Online mode** allows the user to Create/Read/Update/Delete records (CRUD operations) in-memory, interacting directly with a database or with a data frame (just like in Pandas). 
-  **Offline mode**, the user just export the knowledge graph to files.

With Neo4j:
- Online: the user can interact with the Neo4j database, doing the CRUD operations.
- Offline: the user exports files that can be read by a Neo4j engine.
   
With CSV (comma separeted file)
- Online: because it is not implemented, the user cannot read a CSV file, and create, read, update or delete rows.
- Offline: the user can export the knowledge graph in a CSV file.
   
With Pandas:
- Online: all the data is in form of in-memory data frame, so the user can do CRUD operations.
- Offline: the user can export the data in different formats including CSV.

If am I right with the above, I suggest the following.

> 1. Is the above table correct? Does it reflect current capabilities? What does the ideal table look like?

I propose to extend the table with information related to the supported protocols, ordered from the most recommended to the least. For instance, in Pandas, I recommend `.parket`, then `.hdf5` and the last one `.pickle`.
Screenshot from 2024-10-04 12-06-10

> 2. Should we harmonise the settings, and how?

- Mainly, I see we need to define the type we are intended to access, for instance, the database.
- So, if the specification is Offline, details about the database and connection are sufficient.
	
> 3. Which settings should be controlled from the config, which from the Python API (from the user perspective)?

At this point, I cannot suggest which one specifically, but as an example I can assume all the KG formats need different configuration parameters. However, my recommendation at this point is:
   **Config file:**
- File paths.
- Type of database.
- Offline/Online parameter.
- Delimiter, in case of writing CSV files, this could be set by default in the config, but in the Python code we should override this default.

**Python code:**
- Connections to the database, in case i need to access the same type of database (a neo4j for example), but with different connection parameters.

> 4. Which is the simplest and most intuitive approach for the user?
      	
From my perspective, 
1. I should allow the user to decide if the operations will be offline or online:
    - Using the config file.
2. Specific commands based on the type of operation
    - Online (CRUD operations)
        - Connecting to a resource:
            - `connect_to_<neo4j, postgreSQL, SQLite, etc>`,  for instance: `connect_to_neo4j()`, `connect_to_mysql()`.
         - CRUD operations
            -  create_record()
            - read_record()
            - update_record()
            - delete_record()
    - Offline (C operation)
        - `export_to_<.parquet, .json, .csv, Neo4j DB, SQLite DB>`
    
    Basically, we need to explicitly tell to the user that writing into a database is not the same like exporting a file.
	 
> 5. Should we simplify the Python API to only have one method for adding entities/relationships to the KG (as opposed to having add, write, and merge methods)?

What it the purpose of `add`, `write` and `merge` in the context of knowledge graphs? It seems they have different functionalities, if so, reducing them into single method is not a good idea.

> 6. If so, do we (the users) need manual control over the write process in the offline case?

At least if this is a CSV, let the user define the delimiter to use is important, the name of the file, and other considerations.

ecarrenolozano:
I have another question, How do use `admin import` for outputting a Graph? I understand that `admin import` is used for importing data, and this data will be converted into a native Neo4j knowledge graph.

ryxx0811:
> What it the purpose of `add`, `write` and `merge` in the context of knowledge graphs? It seems they have different functionalities, if so, reducing them into single method is not a good idea.

@ecarrenolozano  From my understanding, currently, they should be called for dealing with different modes. (e.g `write` is used in offline mode, since it should export the graph in a file and `add` is responsible for online mode with in_memory). I think @slobentanzer was referring Nils proposal #358（right？）, with `offline` and `dbms` specified in config controlling the flow. Given this, should we simplify the API ?

ecarrenolozano:
Exactly, they have different functions. the purpose of `add` is not the same as `write`. So, in my perspective as an user, I prefer to have separated concepts. How much are we earning by simplifying the API in comparison with having separated functions?

slobentanzer:
Agree with the discussed points in the general sense. Practically and semantically, when it comes to the API, it is more subtle. For instance, we do `add` nodes (to some internal representation) before we `write` them. The separation into two methods is just for explicitly showing the target format. Internally, everything is `add`ed at some point.

There also is no need for an "either/or" decision. We can allow `add` to do everything (even adding both nodes and edges) in simple cases, while still maintaining the set of `<add_nodes(), add_edges(), merge_nodes(), merge_edges(), write_nodes(), write_edges()>` specific methods. We need to teach users how they are used in any case, my main aim is just to make this as self-explanatory as possible.

> I have another question, How do use admin import for outputting a Graph? I understand that admin import is used for importing data, and this data will be converted into a native Neo4j knowledge graph.

@ecarrenolozano I am not sure I understand the question. Admin import is a Neo4j call that imports a KG from specifically formatted files. The BioChatter `write` methods in Neo4j mode write these files, as well as the import call needed to use the Neo4j CLI (using `write_admin_import_call()`),

slobentanzer:
Taking the initial point further, we could also refactor the API to do both adding and writing in two separate steps, like

```
bc.add_nodes(some nodes)
bc.add_edges(some edges)
bc.write_kg() (or .write_kg_to_neo4j())
```

This can become as explicit as we like, but would be a breaking change, so I would not implement this straight away (not in this PR). For this discussion, I would focus only on refactoring that do not break the existing workflow; however we can add methods that can be wrapped by the existing commands in preparation for a larger refactor down the line.

ecarrenolozano:
Exactly, if I am an user working with Pandas (for instance). For me, it is explicit to `add_nodes()`, `add_edges()` (I am correctly assuming it happens in-memory). In addition, if I have to persist the information as files or anything else `write_kg()` is explicit as well.

ecarrenolozano:
> Agree with the discussed points in the general sense. Practically and semantically, when it comes to the API, it is more subtle. For instance, we do `add` nodes (to some internal representation) before we `write` them. The separation into two methods is just for explicitly showing the target format. Internally, everything is `add`ed at some point.
> 
> There also is no need for an "either/or" decision. We can allow `add` to do everything (even adding both nodes and edges) in simple cases, while still maintaining the set of `<add_nodes(), add_edges(), merge_nodes(), merge_edges(), write_nodes(), write_edges()>` specific methods. We need to teach users how they are used in any case, my main aim is just to make this as self-explanatory as possible.
> 
> > I have another question, How do use admin import for outputting a Graph? I understand that admin import is used for importing data, and this data will be converted into a native Neo4j knowledge graph.
> 
> @ecarrenolozano I am not sure I understand the question. Admin import is a Neo4j call that imports a KG from specifically formatted files. The BioChatter `write` methods in Neo4j mode write these files, as well as the import call needed to use the Neo4j CLI (using `write_import_call()`), in some output directory. Executing this import call (shell command) will create a Neo4j DB with the contents of the KG. Is that what you wanted to know? It should be explained in the docs.

It was confusing. Now, I can say that the Offline mode for Neo4J generates two types of files: CSVs and a .sh script that generates the Neo4j graph. I will add that information on the table for reference.

slobentanzer:
@ecarrenolozano true, but it is also explicit: `write` only writes KG files, `write_import_call` only writes the .sh (using the information from the build as to which files there actually are to import, and where they are). These are things the user specifies by using the Python API.

slobentanzer:
Sorry I think there was a GitHub failure, I didn't mean to close this.

ecarrenolozano:
Based on the [OC0001 - NetworkX Output Component (sequence and activity diagram)](https://drive.google.com/file/d/1mNxzEWkAYC6pf-wOf-fwsUfbOg-f5xVM/view?usp=sharing)

**Question:** How is this process related to the parameters in the BioCypher config file? For instance, the `dbms` and `offline/online` mode that is currently under discussion.

ryxx0811:
> Based on the [OC0001 - NetworkX Output Component (sequence and activity diagram)](https://drive.google.com/file/d/1mNxzEWkAYC6pf-wOf-fwsUfbOg-f5xVM/view?usp=sharing)
> 
> **Question:** (@ryxx0811) How is this process related to the parameters in the BioCypher config file? For instance, the `dbms` and `offline/online` mode that is currently under discussion.

Since users only perform add operations, they can decide whether to write to a file or write to database. In this case, I think `offline` is not necessary. The dbms determines the how write functions should work.( in the `write_to_file` , where it could be CSV, Networkx, Neo4j, or the the driver to connect in the `write_to_dbms).`

slobentanzer:
Thanks @ryxx0811, but there are some questions.

Basic premise: we receive collections of tuples (lists or generators) from the adapter; that is the basic representation of nodes and edges in biocypher.

Q1: Why would we pass them to become first a NetworkX object, and then also a dataframe, and then back to the core?

- seems overly complicated
- triples memory footprint (to the original tuples, we add a networkx object, and also dataframes)
- dataframes are not an efficient representation of a graph (that's the reason we created biocypher)

Q2: Why is the same function going from core to networkx, and then from networkx to pandas?

- seems unintuitive for the user
- does it all happen when `add` is called once?

More explanation: the `write` mode exists partly because we can stream data from the adapters. Particularly for large graphs, you can start writing to disk before the end of the collection (default is 1M entries are written per file/batch). In the `add` mode on the other hand, you assume that you can put everything in memory (and thus that your KG fits in memory). For small graphs, this does not make a difference, but for large ones, this suggestion may break usability.

I am somewhat convinced that this configuration of `offline: true` or `false` is not intuitive, more than happy to get rid of that if we can. Not too sure whether we can have just one method of adding/writing nodes and edges. Happy for input.

Also happy for you to take notes, @ecarrenolozano @kpto, whether some of this should go somewhere in the docs to explain biocypher better. :)

ryxx0811:
> Q1: Why would we pass them to become first a NetworkX object, and then also a dataframe, and then back to the core?

When users call the `add_nodes` or `add_edges` methods, the Biocypher core interacts with the NetworkKG class. NetworkxKG class also contains its own add_nodes and add_edges functions, which calls `add_table` in PandasDF class to convert the input data into DataFrame and return it to NetworkxKG. This process inolves only one conversion(from input to Dataframe). Here Dataframe is used to create knowledge graph when `get_kg` is called.

> Q2: Why is the same function going from core to networkx, and then from networkx to pandas?
> 
> * seems unintuitive for the user
> * does it all happen when `add` is called once?

Yes, users only need to call `add_nodes` or `add_edges` in Biocpyher core once. 



> In the `add` mode on the other hand, you assume that you can put everything in memory (and thus that your KG fits in memory).

I think `add` here can still process data in a streaming manner. But users need to call `write_to_file` every time after `add_nodes` or `add_edges`.

ecarrenolozano:
I agree in this case with Sebastian, from an user perspective I prefer:

- use Biocypher to build my KG, it means I could import lots of data. If I am interested in add edges and nodes, we use `bc.add_nodes/add_edges`.

- in case of exceeding memory, internally Biocypher can save the KG, it could be in parquet. And warn the user that this knowledge graph is store due to exceeding memory.

- the user must have the option to convert the Biocypher internal representation to pandas and/or to NetworkX. It doesn't mean that Biocypher will work in pandas or NetworkX, for adding edges or nodes. For instance, by using to_df or to_networkx

- the user persist the knowledge graph as desired, for instance using write_to(). This function must have the capacity to write to disk, access a dB and store the data in there. We can separate concerns and say write_to_file and write_to_db. 

I take advantage of this comment to suggest  the following: put a list of specific requirements that this development should have. They will serve as a base to create test cases.

slobentanzer:
> I take advantage of this comment to suggest the following: put a list of specific requirements that this development should have. They will serve as a base to create test cases.

This is a great suggestion, would you like to implement this? Can be a comment in the discussion we already opened on the output format, right?

Issue #363, opened 2024-07-18
Error importing csv files with Neo4j v.5.21.2


mbaric758:
Problem with importing data with version 5.21.2. The problem seems to occur whenever trying to define a database to import to at the end of the import sequence, for example the name is neo4j here below. Even if the database exists (for example using the name of the default database), there's a `File 'neo4j' doesn't exist` error. Everything works correctly if the name of the database is deleted altogether, the import is then correct, and imports to the default database (neo4j).

In addition to that, in previous versions of Neo4j changing the` initial.dbms.default_database= `name in ` neo4j.conf ` was enough to change the default database and import there by changing the database name at the end of the import sequence, but here it seems to be creating problems with the new version.

For example:

`
neo4j-admin database import full --delimiter=";" --array-delimiter="|" --quote="'" --overwrite-destination=true --skip-bad-relationships=true --skip-duplicate-nodes=true --nodes="/Users/mbaric/ontoweaver1/biocypher-out/20240716155548/Patient-header.csv,/Users/mbaric/ontoweaver1/biocypher-out/20240716155548/Patient-part.*" --nodes="/Users/mbaric/ontoweaver1/biocypher-out/20240716155548/SequenceVariant-header.csv,/Users/mbaric/ontoweaver1/biocypher-out/20240716155548/SequenceVariant-part.*" --relationships="/Users/mbaric/ontoweaver1/biocypher-out/20240716155548/PatientHasVariant-header.csv,/Users/mbaric/ontoweaver1/biocypher-out/20240716155548/PatientHasVariant-part.*" neo4j --verbose`

And this is what I get with --verbose:

```
[...]

Issue #342, opened 2024-04-24
CI: Python 3.9 not on MacOS latest any more, but 3.12 also fails


slobentanzer:

nilskre:
Temporary workaround: Set macos to the last arm version as described [here](https://github.com/actions/setup-python/issues/808).

nilskre:
Current solution: fix the macos runner version to `macos-13`.

The `setup-python` action is now fixed for `macos-14` (at the moment this is `macos-latest`).
But the `[setup-docker-macos-action](https://github.com/douglascamata/setup-docker-macos-action)` will only be updated for macos-15 (see [here](https://github.com/douglascamata/setup-docker-macos-action)).

slobentanzer:
turned off docker-dependent tests in CI for macOS for now, will revisit later. for now, online tests are run on Linux only.

PR #365, opened 2024-07-23
fix(import_call):Database name after import prefix

mbaric758:
Refs: #363

### DIFF
--- a/.github/actions/test/action.yaml
+++ b/.github/actions/test/action.yaml
@@ -11,9 +11,9 @@ runs:
     #    setup docker containers for testing
     #----------------------------------------------
     # currently only available for ubuntu and macos
-    - name: Install Docker
-      uses: douglascamata/setup-docker-macos-action@v1-alpha
-      if: ${{ runner.os == 'macOS' }}
+#    - name: Install Docker
+#      uses: douglascamata/setup-docker-macos-action@v1-alpha
+#      if: ${{ runner.os == 'macOS' }}
     - name: Start Neo4j Docker
       run: docker run --restart always --publish=7474:7474 --publish=7687:7687 --env NEO4J_AUTH=neo4j/your_password_here --env NEO4J_PLUGINS='["apoc"]' --env=NEO4J_ACCEPT_LICENSE_AGREEMENT=yes -d neo4j:${{ inputs.NEO4J_VERSION }}
       shell: bash
--- a/.github/actions/test/action.yaml
+++ b/.github/actions/test/action.yaml
@@ -1,8 +1,8 @@
-name: 'Test and code quality'
-description: 'Run tests and code quality checks'
+name: "Test and code quality"
+description: "Run tests and code quality checks"
 inputs:
   NEO4J_VERSION:
-    description: 'Neo4j version'
+    description: "Neo4j version"
 
 runs:
   using: "composite"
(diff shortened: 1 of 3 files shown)


nilskre (changes requested):
Hi @mbaric758  ,
thanks for investigating and providing this Pull Request!
This leaked, that our current unit test suite lacks the behavior of checking the correct functioning of the import script.
To avoid this in the future: Can you add unit tests for checking this?
Thanks in advance.

mbaric758:
Hi @nilskre I added a test to check whether the import call works, however I think there might be a problem with the `bin/` prefix which is added by default to the import call by the `_get_default_import_call_bin_prefix` function in `_neo4j.py`. The way I usually interact with Neo4j is by adding the path of `neo4j-admin` to `PATH` and `PYTHONPATH` so it's a global variable and the import call works without changing directories to the location of Neo4j installation. 

In case of having to change directories to the Neo4j path before calling the import call, I think this would create problems since pytest has to be called from the biocypher root directory. I didn't change the addition of the `bin/` prefix to the import call in the current PR, so maybe you should also run the test on your system to see if the test works?

slobentanzer:
Hi @mbaric758,
the bin prefix derives from the way that Neo4j handles its folder structure. If you execute a script in the root of a Neo4j DBMS, the neo4j-admin binary is in the bin directory. This is important for the use case of multiple DBMSs on one system. You cannot just put all of them on the path.

Regarding the unit tests: these should not be dependent on the operating system whatsoever. Ideally, we would write tests that test only the validity of the import call, without having to execute it. It is up to us to come up with a method to do this (can be regex or anything else that lets you compare / validate a string). :)

mbaric758:
Thanks @slobentanzer for explaining this ! I understood from the previous comment that the test should run the import call itself to check its functionality, but in that case the `test_construct_import_call` in `test_neo4j.py` should be sufficient for checking the import call structure since only the database name location has changed.

slobentanzer:
I think @nilskre's remark was pointing out that we are not currently considering the bug that you found, i.e. the order of arguments. Maybe we should just add a check for the order then?

mbaric758:
In that case, I modified the existing `test_construct_import_call`, to also check that the database name is now located after the `import` command.

slobentanzer:
@nilskre something's up with the MacOS runners, they seem to fail arbitrarily. Is this related to #342?

nilskre:
> @nilskre something's up with the MacOS runners, they seem to fail arbitrarily. Is this related to #342?

Yes, that is the related issue. I tried one week ago to fix this, by setting the CICD to macos-latest again. The current state is documented there.

Current state summary:
- macos-latest = macos-14 at the moment
- We have set the CICD to use macos-13 to fix the Python setup issue we encountered (until the maintainers have fixed the Python setup action (is now fixed for macos-latest).
- Unfortunately the Docker installation action is not working with macos-14 due to incompatible Apple hardware (see linked documentation in #342). The Docker installation action will only be updated for the future macos-15.
=> Thus I have not found a fast and easy fix for this at the moment.

slobentanzer:
> Thus I have not found a fast and easy fix for this at the moment.

I am not sure I fully get what goes wrong, but it failing arbitrarily is a nuisance. Would it be a solution to remove the Neo4j/Docker stuff from the MacOS runners so they at least run quickly? We could argue that we test the online Neo4j functionality on Linux, and only return it to MacOS if they fix the Apple support / Docker workflow.

nilskre:
Yes, that would be an option 👍

mbaric758:
Hi @nilskre , is there an update on the status of the PR, or are we still waiting for the MacOS testing issue to be resolved ?

nilskre:
Hi @mbaric758 ,
sorry for any inconvenience while waiting for this PR to be resolved.
At the moment I am on vacation and will be back in Biocypher development probably later this year.
To get this done and merged: You could implement the proposed fix to remove the Neo4j Docker tests for Macos from the CICD pipeline (should be sufficient to comment the line where Docker is installed for Macos ( https://github.com/biocypher/biocypher/blob/3de6eca31a29734a7473fdbcae41c71bb3aba678/.github/actions/test/action.yaml#L14).
Thanks in advance!
And then it should be easy for @slobentanzer to quickly review and merge.

slobentanzer:
Hi @mbaric758,
sorry for the delay, as Nils explained, he is currently off and I was busy with submitting the latest paper. I fixed the CI configuration and will review the PR soon.

slobentanzer (approved):
Looks good, many thanks @mbaric758 :)

Bumping patch version

Issue #305, opened 2024-01-04
Resource class and downloader: cache API requests


slobentanzer:
We can download files, compressed files, lists of files; a great addition would be to cache the results of API requests.

PR #369, opened 2024-08-13
add function for caching API requests and refine the args description

ryxx0811:
related to #305

### DIFF
--- a/biocypher/_get.py
+++ b/biocypher/_get.py
@@ -101,19 +101,21 @@ class Downloader:
         self.cache_dict = self._load_cache_dict()
 
     # download function that accepts a resource or a list of resources
-    def download(self, *downloads: Union[Resource, APIRequest]):
+    def download(self, *resources: Union[Resource, APIRequest]):
         """
-        Download one or multiple resources, APIrequest, or both.
+        Download one or multiple files, APIrequest, or both.
 
         Args:
-            downloads (Resource or APIRequest): The resource(s) or API request(s) to download.
+            resources (Resource or APIRequest): The resource(s), i.e., file(s)
+                or API request(s), to download.
 
         Returns:
-            list[str]: The path or paths to the downloaded resource(s) or API request(s).
+            list[str]: The path or paths to the downloaded resource(s) or API
+                request(s).
         """
         paths = []
-        for download in downloads:
-            paths.append(self._download_or_cache(download))
+        for resource in resources:
+            paths.append(self._download_or_cache(resource))
 
         # flatten list if it is nested
         if is_nested(paths):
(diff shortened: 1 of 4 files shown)


biocypher/_get.py
```diff
@@ -335,6 +390,71 @@ def _update_cache_record(self, resource: Resource):
...
+        else:
+            with open(path, "r") as file:
+                logger.info(
+                    f"Attempting to load {api.name} API request data from file: {path}."
+                )
+                response_data = json.load(file)["api_request"]
+
+        return response_data
+
+    def _cache_api_request(self, response_data, api: APIRequest):
```
slobentanzer, line 425: Where is this method used, why is it here?

slobentanzer (changes requested):
Hi @ryxx0811,
many thanks, the code already looks very good. However, I have some questions regarding the functionality. In general, it is advisable to describe the changes you made in the PR comment, so that any reviewer has an easier time understanding the purpose.

I see that you created an `APIRequest` class analogous to the `Resource`, and private methods for getting and caching, but I don't see how it is integrated in the workflow. We have the `Downloader` class with the main `download` method, and that, so far, has no connection to the `APIRequest`. Can you think about how the user would interact with the `Downloader` API to download-or-cache an API request in the same way that we do with file downloads?

Please try to make the final API as simple as possible for the user (regardless of downloading files, API requests, or a mix), and align with the current functionality (described [here](https://biocypher.org/api.html#download-and-cache-functionality)). It would also be great if you would then record the updated functionality with usage instructions in those API docs, which are generated from [here](https://github.com/biocypher/biocypher/blob/main/docs/api.rst).

Please also make sure to test all of the new functionality, not just the private methods. Any questions, just let me know. :)

biocypher/_get.py
ryxx0811: Thank you for the feedback.Initially, my idea was to create a separate function for getting and caching API requests respectively...so according to what you mentioned,  from the user's perspective, the download method should not only handle the downloading of resources but also be the caching API requests, correct?

slobentanzer: Yeah, the `download` method should not care whether it's `Resource` or `APIRequest`. It should be able to accept arbitrary collections (such as a list) of `Resource`s and `APIRequest`s.

You created the download and cache functions as private methods (leading underscore), which usually means that you do not want the user to interact with them, right?

ryxx0811:
The `download` method accepts the _downloads_ variable, which can be either a single or multiple resources, API requests, or a mix of both.
The `_download_api_request` method is created to fetch and cache API requests, and it will be called by the `download` method.
The `_download_resource` function was modified to return a list to avoid issues when `download` method handles multiple resources, where one resource has a single file and another has multiple. This prevents paths from being incorrectly split into individual characters by the `is_nested` function.(see `test_path_splitted` method).

slobentanzer:
@ryxx0811 I recommend running the pre-commit hooks, otherwise the CI may fail at that step (in this case, `black` didn't pass)

biocypher/_get.py
```diff
@@ -80,74 +101,79 @@ def __init__(self, cache_dir: Optional[str] = None) -> None:
...
-        if os.path.exists(resource_cache_path) and os.path.isdir(
-            resource_cache_path
-        ):
-            shutil.rmtree(resource_cache_path)
+    def _delete_expired_cache(self, download: Union[Resource, APIRequest]):
+        cache_path = self.cache_dir + "/" + download.name
+        if os.path.exists(cache_path) and os.path.isdir(cache_path):
+            shutil.rmtree(cache_path)
 
     def _download_resource(self, cache, resource):
```
slobentanzer, line 178: Types should be hinted here, not clear if resource is only Resource or also APIRequest

biocypher/_get.py
```diff
@@ -61,6 +63,25 @@ def __init__(
         self.is_dir = is_dir
 
 
+class APIRequest:
```
slobentanzer, line 66: Could this class not inherit from `Resource`? An `APIRequest` *is* a resource in the sense of our module, right? This would simplify type hinting, and we would only have to check which class is passed when deciding between pooch and requests.

test/test_get.py
```diff
@@ -196,37 +210,147 @@ def test_download_zip_and_expiration():
         "https://github.com/biocypher/biocypher/raw/main/test/test_CSVs.zip",
         lifetime=7,
     )
-    paths = downloader.download(resource)
+    api = APIRequest(
```
slobentanzer, line 213: Why is this test included in `test_download_zip_and_expiration`? I think the test of API request functionality deserves its own battery of tests. Please move to separate functions.

biocypher/_get.py
```diff
@@ -80,74 +101,79 @@ def __init__(self, cache_dir: Optional[str] = None) -> None:
...
 
         if expired or not cache:
-            self._delete_expired_resource_cache(resource)
-            logger.info(f"Asking for download of {resource.name}.")
-            paths = self._download_resource(cache, resource)
+            self._delete_expired_cache(download)
+            if isinstance(download, Resource):
+                logger.info(f"Asking for download of resource {download.name}.")
+                paths = self._download_resource(cache, download)
+            else:
```
slobentanzer, line 142: I would check here for the instance of the API class. Not safe to assume that everything that is not a `Resource` is an `APIRequest`. Better to test explicitly.

test/test_get.py
```diff
@@ -196,37 +210,147 @@ def test_download_zip_and_expiration():
...
+
+@pytest.mark.parametrize(
+    "downloader",
+    [
+        "downloader_without_specified_cache_dir",
+        "downloader_with_specified_cache_dir",
+    ],
+    indirect=True,
+)
+def test_path_splitted(downloader):
```
slobentanzer, line 323: What is the purpose of this test, what does "path splitted" mean?

test/test_get.py
```diff
@@ -196,37 +210,147 @@ def test_download_zip_and_expiration():
...
+            "https://github.com/biocypher/biocypher/raw/main/biocypher/_config/test_schema_config_disconnected.yaml",
+        ],
+    )
+
+    paths = downloader.download(resourc1, resource2)
+    assert isinstance(paths, list)
+    assert len(paths) == 3  # if len(paths) == 3, then not splitted
+
+
+def test_download_with_parameter():
```
slobentanzer, line 342: Looks like indeed the download works now. However, looking at the returned path, the parameter (`?download=1`) is included in the file name, which I am not sure we want. Should we remove parameters from downloaded files? When downloading manually, we would just receive a `CollecTRI_source.tsv`.

slobentanzer (changes requested):
Hi @ryxx0811, looking very good. I have some suggestions for improvements, please see the individual comments.

Another request: please update the documentation of the `_get.py` API as discussed previously to allow users to see all available functions we provide. It's in `docs/api.rst`, line 183.

biocypher/_get.py
slobentanzer: We could even make two distinct classes, `FileDownload` and `APIRequest`, which both inherit from an abstract base class `Resource`, to make the distinction even more clear. Thoughts about that?

ryxx0811: @slobentanzer yes, you are right. I will make two classes inherit from abstract `Resource`：）

test/test_get.py
ryxx0811: @slobentanzer When `download` method handles multiple resources, where one resource has a single file and another has multiple, the path(the path for resource1) will be split into string characters...so I modified the `_download_resource` to return a list.In the test, if the length pf paths is 3(one for resource1 and two from resource2), then path for resource1 is returned correctly.

slobentanzer (approved):
Good job, this now seems to do everything we wanted. :) Will merge and bump the minor version to `0.6.0`.

slobentanzer (approved):

Issue #296, opened 2023-11-27
Resource download: currently does not allow params (e.g. `?download=1` of Zenodo)


slobentanzer:
Could add parameter to resource url spec

PR #371, opened 2024-08-28
Related to #296:Resource download with parameter

ryxx0811:
issue #296: add a test method `test_download_with_parameter():` downloading resource with query parameter.


slobentanzer:
@ryxx0811 isn't this also in the other PR? We can work on it there, no need to have two PRs for similar work. I think we can close this.

ryxx0811:
@slobentanzer

PR #379, opened 2024-10-10
Folder for drawio diagrams 

ecarrenolozano:
**Current**
This pull request do the following:
- remove all the .svg images from the /docs folder.
- remove the folder /docs/diagrams

**Before:**
 	~~The world is flat.~~
~~This PR comes from the Discussion (#377)~~ 

~~Tasks involved in this issue~~

- ~~Creation of a folder in `/docs/diagrams`~~
    - ~~**Purpose:** Centralize all the diagrams for the BioCypher repository.~~
- ~~Architecture diagrams (Architecture_BioCypher.drawio) have been added.~~
    - ~~format: .drawio.svg~~
    - ~~The original file (with all diagrams) has been split.~~


.pre-commit-config.yaml
```diff
@@ -4,8 +4,8 @@ fail_fast: false
 default_language_version:
     python: python3
 default_stages:
--   commit
--   push
+-   pre-commit
```
kpto, line 7: Please explain the change here and be more descriptive on your commit message, your commit message should talks about this part as well.

slobentanzer:
Agree with the size issue, generally. If we do svg, it's saved as an image. Which is better for referencing it in discussions (it is rendered on GitHub directly), but worse in terms of storage. I didn't know draw.io had a GitHub plugin, will check out the request @kpto opened. Maybe that's the solution.

slobentanzer:
I installed the app, please check if it works for our purposes

slobentanzer:
> Working perfectly, can open a diagram using draw.io directly from the repo.

@kpto can you explain how (or link to docs)?

kpto:
> > Working perfectly, can open a diagram using draw.io directly from the repo.
> 
> @kpto can you explain how (or link to docs)?

You must use the web app [https://app.diagrams.net/](url)

The UI is pretty self explanatory, just open the link above, choose GitHub, choose the repo and a branch, then the file you want to edit.

slobentanzer:
> changes on documentary diagrams shouldn't trigger tests, we will need to fine tune the configuration to exclude certain directories.

yes. we can just exclude the `docs` dir.

> You must use the web app

So I guess I can just keep using the VSCode plugin, which offers the exact same without having to go to a third party web app. Which is good (anybody can use whichever app they want).

I don't fully understand the size argument; what format should we save the files as? Does it actually make a difference?

slobentanzer:
No, I think the test fail is unrelated (some pre-commit thing) and also tests shouldn't run when we only update the docs. This is a separate issue, I would suggest only doing the diagram changes in this PR, and move the rest to a different issue / PR.

slobentanzer:
@kpto there is one complication that I can think of, which is that we use doctest to verify the code in the documentation is still correct and current in comparison with the real code. There may be some entanglement there.

.pre-commit-config.yaml
slobentanzer: I would even suggest to revert this and make the subject of a separate issue and PR. I have not thought deeply about the pre-commit setup, but this is certainly something we can discuss (in combination with, for instance, whether documentation changes should trigger the test).

ecarrenolozano:
> > So I guess I can just keep using the VSCode plugin, which offers the exact same without having to go to a third party web app. Which is good (anybody can use whichever app they want).
> 
> Yes you are good.
> 
> > I don't fully understand the size argument; what format should we save the files as? Does it actually make a difference?
> 
> I don't think that's an argument but merely an observation that a file sized too large leads to test fail. But regarding the format to use, I recommend .drawio.svg, vector based, embeddable, native draw.io information kept, no reason not to choose, except that it somehow leads to test fail problem which requires investigation. I don't think a test should fail because of this?

When I made the test, it failed for the size. That's why it was mentioned in the comments. And that is why I suggested to store the images in .drawio or .drawio.svg

slobentanzer:
> When I made the test, it failed for the size.

You also changed the YAML configuration, which I think we should revert and discuss in a separate issue / PR, see the other code comment. This PR should only be about the diagram folder.

kpto:
> > When I made the test, it failed for the size.
> 
> You also changed the YAML configuration, which I think we should revert and discuss in a separate issue / PR, see the other code comment. This PR should only be about the diagram folder.

Totally agree, I wanted to mention this in the beginning actually but worried that I might go too far on enforcing the rules. Though, I must say that sometimes breaking the rules really remove a lot of hurdle so I am fine with it occasionally. As long as the mentality is set right, I don't want to be too strict on others.

.pre-commit-config.yaml
kpto: Yes, see [https://github.com/biocypher/biocypher/pull/379#issuecomment-2407561946](url)

slobentanzer:
> Totally agree, I wanted to mention this in the beginning actually but worried that I might go too far on enforcing the rules. Though, I must say that sometimes breaking the rules really remove a lot of hurdle so I am fine with it occasionally, as long as the mentality is set right. I don't want to be too strict on others.

@kpto, absolutely, balance is needed. However, since we are just starting to work together, it can be worth taking some time to explain some ground rules in more detail, rather than some time later. But in essence you are right, it depends on the circumstance. (makes me also think doing these "user manuals" for ourselves may be valuable to do some time in the near future)

kpto:
@slobentanzer Have you reviewed the diagrams and confirm their correctness? The mission part is still outside my knowledge so I will rely on you on that part. I will need to have a thorough code reading to verify the architectural diagrams which will takes some time. I already think the configuration part was not drawn correctly, one of my architectural concern is exactly that the configuration layer is not shared by modules.

@ecarrenolozano I don't understand the file naming, what do the codes prefixing each file mean?

slobentanzer:
@kpto the architecture diagrams are not correct or complete (briefly skimmed). The PR is just to introduce a folder for them in order to be able to use them in discussion #377, so they don't need to be fully correct. We can continue improving them as we discuss what the architecture should be.

@ecarrenolozano have you created an issue for the pre-commit changes you wanted to make? If everything is in order, we can merge this PR to get on with the architecture discussion.

ecarrenolozano:
> @slobentanzer Have you reviewed the diagrams and confirm their correctness? The mission part is still outside my knowledge so I will rely on you on that part. I will need to have a thorough code reading to verify the architectural diagrams which will takes some time. I already think the configuration part was not drawn correctly, one of my architectural concern is exactly that the configuration layer is not shared by modules.
> 
> @ecarrenolozano I don't understand the file naming, what do the codes prefixing each file mean?

We started the diagram process recently, based on the documentation and how the BioCypher works (not diving into nitty-gritty details for now). So, this is just a starting point to formalize them. 

The file convention is based on Legacy Diagrams, Migration Diagrams, Adapter Components, and Output Components. Each diagram needs a unique code, so please I am open to suggestions related to the names but they should keep a unique code.

kpto:
> @kpto the architecture diagrams are not correct or complete (briefly skimmed). The PR is just to introduce a folder for them in order to be able to use them in discussion #377, so they don't need to be fully correct. We can continue improving them as we discuss what the architecture should be.

I don't agree with knowingly commit wrong information into a repo but I am flexible so I am fine, but I still want the file naming sticks to the established convention as much as possible. From what I see in the docs directory the naming is a mixture of snake case and kebab case so please use either one of them. In the future we should have a consensus on naming and rename those files.

ecarrenolozano:
> > @slobentanzer Have you reviewed the diagrams and confirm their correctness? The mission part is still outside my knowledge so I will rely on you on that part. I will need to have a thorough code reading to verify the architectural diagrams which will takes some time. I already think the configuration part was not drawn correctly, one of my architectural concern is exactly that the configuration layer is not shared by modules.
> > @ecarrenolozano I don't understand the file naming, what do the codes prefixing each file mean?
> 
> We started the diagram process recently, based on the documentation and how the BioCypher works (not diving into nitty-gritty details for now). So, this is just a starting point to formalize them.
> 
> The file convention is based on Legacy Diagrams, Migration Diagrams, Adapter Components, and Output Components. Each diagram needs a unique code, so please I am open to suggestions related to the names but they should keep a unique code.

@slobentanzer you can merge this PR. Then, I could create a new PR request exclusively to discuss the suggestion of migration for the pre-commit-config.yaml file.

kpto:
> The file convention is based on Legacy Diagrams, Migration Diagrams, Adapter Components, and Output Components. Each diagram needs a unique code, so please I am open to suggestions related to the names but they should keep a unique code.

Why? I am not familiar with such a practice, what are the codes for when your file names are already descriptive?

kpto:
> @slobentanzer you can merge this PR. Then, I could create a new PR request exclusively to discuss the suggestion of migration for the pre-commit-config.yaml file.

Shouldn't this be an issue or discussion?

ecarrenolozano:
> > @slobentanzer you can merge this PR. Then, I could create a new PR request exclusively to discuss the suggestion of migration for the pre-commit-config.yaml file.
> 
> Shouldn't this be an issue or discussion?

This is the message : 

> [WARNING] top-level `default_stages` uses deprecated stage names (commit, push) which will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.

Issue, because it will be removed in the future and it can cause an issue in the future.

slobentanzer:
Re the „wrongness“ of the diagrams: I don’t think it applies this strictly here. Uploaded diagrams could also be suggestions for the refactor, so it is not really possible to determine whether they are correct. 

Re the identifier system: I also don’t see the point in using an arbitrary system that does not carry meaning. For personal use, sure, but as soon as multiple contributors collaborate, this creates an explanatory overhead. I think the uniqueness of the diagrams is guaranteed by the descriptive naming and the versioning provided by GitHub. What I would like as an organising principle is some sort of structured index, like a table of contents in markdown that links the diagrams together. Could be an .md file in the `diagrams` folder. Happy for suggestions of more elegant solutions though.

ecarrenolozano:
> > The codes contribute to organizing and indexing the diagrams. It is a common practice in Systems and software engineering.
> 
> This is really the first time I am hearing this, can you point me to some references or give me a few keywords for what to search on Google to get relevant information? This seems really arbitrary to me like, the number of leading zeros?
> 
> @slobentanzer What do you think about this? Are you familiar with this practice?

TL;DR
- **Codes:** we can remove the codes, any suggestions for names other than that?
- **Structured index:** I suggest just using Markdown and that's it. On the other hand, a Wiki could be another idea, but we don't want to over-complicate things too much do we?

Guys, I don't have a concrete example because it depends on the convention defined in the organization. Usually, organizations have diagramming tools that generate codes in company with prefixes. As, this project is really small, we can do without codes. However, in case you work with projects where the number of diagrams scales, you will probably find this situation. Readings I can perhaps recommend are the C4 specifications, Archimate, TOGAF, and look for Software Engineering standards (IEEE 1471 among others). All of them give different recommendations for software architecture aspects, which in the end end up being decisions of the team or company.

kpto:
> Re the „wrongness“ of the diagrams: I don’t think it applies this strictly here. Uploaded diagrams could also be suggestions for the refactor, so it is not really possible to determine whether they are correct.

@slobentanzer While I am fine with it right now, but I feel the necessity to remind you that this PR is committing changes to the `docs` directory in `main` repo. With the `prod` branch being absent, this branch becomes the branch that holds the latest, confirmed shape of the project. I am fine with it right now because the changes will not be published automatically but if we have a continuous delivery set up in the future, this must be prohibited, even though the diagram itself will not be disclosed due to it being not referenced, the mentality matters. What you said is right though if this PR is committing to a `dev` branch.

To give you a more practical reason, imagine if we somehow need to freeze our software due to a major refactoring, the `main` branch will be a convenient place to pick a state to freeze. If incorrect information is randomly introduced we will have trouble tracing back to a commit which we are certain that it is stable to serve the public for a long term.

Or if you actually mean a place to store ideas, then it would not be `docs` directory but something else, or even a different repo.

kpto:
> > > The codes contribute to organizing and indexing the diagrams. It is a common practice in Systems and software engineering.
> > 
> > 
> > This is really the first time I am hearing this, can you point me to some references or give me a few keywords for what to search on Google to get relevant information? This seems really arbitrary to me like, the number of leading zeros?
> > @slobentanzer What do you think about this? Are you familiar with this practice?
> 
> TL;DR
> 
> * **Codes:** we can remove the codes, any suggestions for names other than that?
> * **Structured index:** I suggest just using a simple Markdown file. On the other hand, a Wiki could be another idea, but we don't want to over-complicate things too much do we?
> 
> Guys, I don't have a concrete example because it depends on the convention defined in the organization. Usually, organizations have diagramming tools that generate codes in company with prefixes. As, this project is really small, we can do without codes. However, in case you work with projects where the number of diagrams scales, you will probably find this situation. Readings I can perhaps recommend are the C4 specifications, Archimate, TOGAF, and look for Software Engineering standards (IEEE 1471 among others). All of them give different recommendations for software architecture aspects, which in the end end up being decisions of the team or company.

I suggest we just merely give a descriptive name in snake case, I am not even sure an index is needed.

slobentanzer:
I like the idea of using the wiki; we don't use it at the moment, and GitHub can render simple diagrams in mermaid. This would also solve the question of how to display the (non-mermaid) images, as I think GitHub wiki also can render any image.

In addition, it will also be a separate repository, which is another comment that I generally agree with. Should we go ahead with the wiki? All we'd need to do is to have a start page explaining that the wiki is developer-focussed, not for the typical user.

slobentanzer:
Created a wiki home page, the repo is https://github.com/biocypher/biocypher.wiki.git. I suppose you should be able to collaborate on that, right?

slobentanzer:
Made an example page with a mermaid diagram, I think this is very useful: https://github.com/biocypher/biocypher/wiki/Example-diagram

slobentanzer (approved):
Great, thanks!

PR #403, opened 2025-01-14
Dev-networkx-ruff

slobentanzer:
This PR is to
- introduce a preliminary version of the in-memory behaviour for KGs; this is supposed to be generalised soon. For now, this means that we can create in-memory versions of a KG either as Pandas or as NetworkX objects in a dedicated fashion. There is no generic internal representation yet, but this is supposed to be the subject of a new draft/issue/PR cycle.
- migrate the code linting from a composite pre-commit setup to ruff. Full ruff checks are implemented at the developer level (use your IDE configuration to execute); pre-commit checks in CI are run using a minimal set of errors (because otherwise we'd have to fix thousands of small things). I propose to fix all these minor warnings consecutively as we work on other aspects of the code.

Follow-up work: 
- draft an issue / user stories for the generic in-memory representation with the goal of implementing through a PR next. I left `TODO`s in the code to identify some of the locations that need improvement.
- further streamline the API from the user's perspective. Probably implement a simpler configuration to select between `online`, `offline`, and `in-memory` use cases, including intuitive Python methods of the BioCypher core.


.github/actions/install/action.yaml
```diff
@@ -3,7 +3,7 @@ description: "Install dependencies"
 inputs:
   PYTHON_VERSION:
     description: "Python version"
-    default: "3.10"
+    default: "3.11"
```
ecarrenolozano, line 6: In previous commits it is mentioned that the minimum version should be 3.10. Do we have a reason to put the default as 3.11 here?

.github/actions/setup/action.yaml
```diff
@@ -3,7 +3,7 @@ description: "Setup Python and Poetry"
 inputs:
   PYTHON_VERSION:
     description: "Python version"
-    default: "3.10"
+    default: "3.11"
```
ecarrenolozano, line 6: In previous commits it is mentioned that the minimum version should be 3.10. Do we have a reason to put the default as 3.11 here?

.github/actions/install/action.yaml
slobentanzer: in short, to be future-proof (shouldn't matter for deployment, this is not the test matrix)

slobentanzer:
Hi @ecarrenolozano, do you have time to review this PR (regarding the introduced functionality)?

ecarrenolozano (approved):
**All the changes are approved.** 
Initially, the tests failed because Ruff adjusted the formatting, and those changes were not automatically committed. The issue was resolved by running Ruff locally, committing the formatting changes, and then re-running the tests.

By the time of this PR, all tests have passed successfully.

PR #405, opened 2025-01-16
Fix neo4j CSV output not escaping quote character and cause import error

kpto:
My recent customer support brought me back to this issue. While this is not the only reason for his trouble but this certainly complicated his problem so I decided to fix it now. This PR detects and escapes quote character appears in a field to prevent import error due to quote character conflict and the subsequent incorrect splitting of a CSV row.

Before fix:
`5'-ABC-3'`, `5'-DEF-3'` results into `'` `5'-ABC-3'|5'-DEF-3'` `'` which cannot be parsed correctly, following neo4j documentation, quote character could be escaped by doubling it so now it becomes `'` `5''-ABC-3''|5''-DEF-3''` `'`

Tested single field and array field to confirm the output could be imported correctly.

Only neo4j is fixed, RDF and Protgres remain unfixed, their behaviour is unchanged as I don't want to invest time on them for now. I still tested the ability to output these two formats to confirm there is no regression.

### DIFF
--- a/biocypher/output/write/_batch_writer.py
+++ b/biocypher/output/write/_batch_writer.py
@@ -19,6 +19,16 @@ class _BatchWriter(_Writer, ABC):
     """Abstract batch writer class"""
 
     @abstractmethod
+    def _quote_string(self, value: str) -> str:
+        """
+        Abstract method to quote a string. Escaping is handled by the database-specific writer.
+        """
+
+        raise NotImplementedError(
+            "Database writer must override '_quote_string'"
+        )
+
+    @abstractmethod
     def _get_default_import_call_bin_prefix(self):
         """
         Abstract method to provide the default string for the import call bin prefix.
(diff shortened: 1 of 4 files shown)


slobentanzer (changes requested):
@kpto great, many thanks. Fully agree on the content. Could you please introduce tests that inform about this edge case and test the logic behind the improvements?

Ideally would test a bunch of hypothetical scenarios of different content and quote choices; in these cases I liked using the [hypothesis](https://hypothesis.readthedocs.io/en/latest/) framework previously (it is already a dev dependency). I feel this is important particularly in this case, as quoting is a recurring issue for many users, especially with Neo4j.

slobentanzer:
Bonus task: if you feel up for it, I would also be very interested in how we can disentangle what Python does vs what the operating system or Java (in the case of Neo4j) do to the characters. I only have a rudimentary understanding of the processes behind the scenes of character processing, and it would be great to get some additional insight, tests, and general recommendations. Not a must for this PR though, just mentioning it because it is a recurring topic. Windows in particular is a repeat offender.

kpto:
Since the logic is rather simple, the verification logic in a test case will be identical to the tested logic and thus the test perhaps does not offer much value, I reckon that a unit test for this is not necessary but instead a system test should be incorporated in the future to truly test the validity of the output against neo4j and other databases.

slobentanzer (approved):
many thanks, added an issue to keep track (#406). will merge this and bump patch.

PR #410, opened 2025-02-03
feat(neo4j): allow several labels sorting methods

jdreo:
Keep the same default as the previous implementation (as "Alphabetical"), for retro-compatibility but I would have found it cleaner to use "Ascending".


slobentanzer (approved):
@jdreo many thanks for the PR! Agree with your assessment and changed to make ascending the default. Since we will be bumping minor version, that should leave enough room for any implementation that may depend on ordering alphabetically to catch up and change the setting. Merging :)

PR #425, opened 2025-03-06
Mkdocs-migration

slobentanzer:
Switch from Sphinx to mkdocs material. The resulting page is already live at biocypher.org.

Allow combination of individual repos under one monorepo (`documentation`), using mkdocs.nav.yaml (which is a separate navigation file). The main build can still be accessible via the github.io link as a single docs page.

Introduce many layout, content, and structural changes.

Add documentation versioning using mike (select version using dropdown). Fairly sure this will not work with the monorepo, again leaving this functionality for the legacy github.io page. I have manually pushed the 0.9.0 version to the gh-pages branch as an initial update.

The docs build workflow has been updated to use mike and respect the versions, but it remains to be tested (can only do that once merged).


docs/index.md
```diff
@@ -0,0 +1,35 @@
...
+
+!!! tip "Hot Topics"
+
+    BioCypher is the simplest way to create an AI-enabled knowledge graph for biomedical (or other) tasks. Check our [BioChatter](https://biochatter.org) documentation for more information.
+
+    We have also recently published a perspective on connecting knowledge and machine learning to enable causal reasoning in biomedicine, with a particular focus on the currently emerging "foundation models." You can read it [here](https://www.embopress.org/doi/full/10.1038/s44320-024-00041-w).
+
+## New to Knowledge Graphs?
+
+If you’re new to knowledge graphs and want to familiarise with the concepts that drive BioCypher, we recommend to check out the graphical abstract below and read [our paper](https://www.nature.com/articles/s41587-023-01848-y) (self-archived version [here on Zenodo](https://zenodo.org/records/10320714), online version [at this https link](https://biocypher.github.io/biocypher-paper/))!
```
ecarrenolozano, line 30: In my opinion, we should change this line: ...we recommend to check out the graphical abstract below... the graphical abstract is above.

ecarrenolozano (commented):
In my opinion, it looks well-organized. Here are a few minor details:
1. The favicon is not rendering.
2. The timeline disappeared.
3. The images look a bit off during the transition between light and dark mode. This could be a potential area for future improvement in the overall look and feel.

docs/index.md
ecarrenolozano: My revision for the BioCypher Ecosystem webpage:

## General

1. I suggest removing the GitHub link to biocypher/documentation (top right of the webpage).

2. I recommend creating a consistent color palette for the organization (umbrella). Child projects can inherit these colors with variations to maintain a unified brand identity. For example, Microsoft Azure serves as an umbrella with visually aligned child products, as seen in this [image](https://cdn-dynmedia-1.microsoft.com/is/image/microsoftcorp/products-hero?resMode=sharp2&op_usm=1.5,0.65,15,0&wid=5760&qlt=100&fit=constrain).

3. Instead of rendering each child project directly on the umbrella webpage, we should use the page to link to each project separately.

## About

1. We definitely need a more visually appealing landing page with an image that illustrates how the projects fit together under a larger initiative (umbrella). Here’s An [incomplete example](https://marvel-b1-cdn.bc0a.com/f00000000151999/www.teradata.de/getmedia/df1ac391-5756-4e7a-b88f-36a5e36f8793/VantageCloud-v2.png?origin=fd) of the type of figure we could use. This should be complemented with the logo and name of each tool.

## Sponsors

1. It would be great to include grid boxes displaying the logos of each sponsor, similar to [this example](https://pytorch.org/foundation).

2. I am unsure whether GitHub, Docker, and Zulip should be listed on this page. If they are included, we could consider reaching out to them in the future to officially showcase the project, as seen [here](https://www.youtube.com/watch?v=pmoG7pLpAAU).

3. Alternatively, we could simply mention that we rely on the following open-source tools to support this project: GitHub, Docker, and Zulip.

## Community

1. We need to update the categories for bug reports and enhancements to align them with the templates we have created.

slobentanzer: Hi @ecarrenolozano,
> I recommend creating a consistent color palette for the organization

didn't you want to do that? :)

> We definitely need a more visually appealing landing page with an image that illustrates how the projects fit together under a larger initiative (umbrella).

can you make concrete recommendations for biocypher? how should it look?

> I am unsure whether GitHub, Docker, and Zulip should be listed on this page.

why are you unsure?

slobentanzer: Hi @ecarrenolozano,
Zulip sponsor our channel (enterprise version), Docker sponsor our Dockerhub (also enterprise) and GitHub sponsor pro-accounts for educational reasons. Thus, we do get services that are normally paid by all of them.

We don't get an enterprise service for free from Neo4j (yet), so the current way of crediting the sponsors is correct, IMO.

Issue #433, opened 2025-04-01
[BUG]: Filename too long Error in case of URL with multiple query parameters


vd-dragan21:
### Source

poetry

### Python version

3.11

### BioCypher version

v.0.9.2

### OS platform and distribution

Ubuntu

### Describe the bug

When I submit an API request with multiple query parameters that are joint in the URL using APIRequest class from Biocypher, I get the Error:

```
API request failed: [Errno 36] File name too long: '/tmp/tmpl99jzyz2/iedb_substring_8caf79be0a786e74a85824ad281fd65d/epitope_search?or=(linear_sequence.ilike.*LQPFPQPELPYPQPQ*,linear_sequence.ilike.*VLQAVGACV*,linear_sequence.ilike.*LPEPLPQGQLTAY*,linear_sequence.ilike.*LDRLNQLES*,linear_sequence.ilike.*NAITNAKII*,linear_sequence.ilike.*YTACSHAAV*,linear_sequence.ilike.*RFYKTLRAEQASQ*,linear_sequence.ilike.*VMVELVAEL*,linear_sequence.ilike.*PSDKHIKEYLNKIQN*,linear_sequence.ilike.*HSKKKCDEL*,linear_sequence.ilike.*ELRRKMMYM*,linear_sequence.ilike.json'
````

Example of request:
```
epitopes = ['QQDKHYDLSY', 'FLKEQGGL', 'EGHLKRNIV', 'GPGMKARVL', 'LMHCQTTL', 'FEAQKAKANKAVDG', 'PQQPFPQPEQPFP', 'RLLMLFAKDVVSRN', 'NDLAVIPAGVVHNWD', 'VHFFKNIVTPRTPG', 'KLVALVINAV', 'SPMRSVLL', 'PADPLAFFSSAIKGGGGSLV', 'FPQGEAREL', 'QPFPQPEQPFP']

conditions = [f"linear_sequence.ilike.*{e}*" for e in epitopes]
check = f"or=({','.join(conditions)})"
base_url = "https://query-api.iedb.org/epitope_search"
url = f"{base_url}?{check}&select=structure_id,structure_descriptions,linear_sequence&order=structure_id"

iedb_request = APIRequest(
    name="iedb_substring_matches",
    url_s=[url],
    lifetime=30,
)

bc = BioCypher()
paths = bc.download(iedb_request)
```

### Relevant LOG output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #434, opened 2025-04-01
Added a method to trim long filenames for the download functions

vd-dragan21:
Fixing issue #433 + small typo fix in api.rst

### DIFF
--- a/biocypher/_get.py
+++ b/biocypher/_get.py
@@ -402,13 +402,13 @@ class Downloader:
         with open(self.cache_file, "w") as f:
             json.dump(self.cache_dict, f, default=str)
 
-    def _trim_filename(self, url: str, max_length: int = 100) -> str:
+    def _trim_filename(self, url: str, max_length: int = 150) -> str:
         """
         Create a trimmed filename from a URL that won't exceed filesystem limits (e.g. for URLs with multiple query parameters)
         
         Args:
             url (str): The URL to generate a filename from
-            max_length (int): Maximum filename length (default: 250)
+            max_length (int): Maximum filename length (default: 150)
             
         Returns:
             str: A valid filename derived from the URL, trimmed if necessary
--- a/biocypher/_get.py
+++ b/biocypher/_get.py
@@ -414,9 +414,9 @@ class Downloader:
             str: A valid filename derived from the URL, trimmed if necessary
         """
         # Extract the filename from the URL
-        fname = url[url.rfind("/") + 1:]
+        fname = url[url.rfind("/") + 1 :]
 
-            # Remove query parameters if present
+        # Remove query parameters if present
         if "?" in fname:
             fname = fname.split("?")[0]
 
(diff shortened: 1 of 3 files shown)


slobentanzer:
@vd-dragan21 many thanks, and congrats on your first PR 🎉

You can request me as a reviewer, that's a more formal way of letting me know I should review the PR. It also helps if you still want to change something and tell me to hold off on the review, then you can only request a review once you're ready.

PS: I just requested Copilot as a reviewer as that's the first time I've seen that suggested, let's see what it has to say.

slobentanzer (approved):
Looks solid; I added another test for the `FileDownload` application, some more tests wouldn't have hurt ;)

Since we want to have this functionality on `main` quickly, I'll leave it at that for this PR; but maybe you can prepare another one with the following:
- more exhaustive tests (e.g., what happens when we provide lists, for instance with separate files that have the same name but different suffix, as can happen)
- Copilot asked about preserving the file suffix: should we do that? Separate off the file suffix and add it back to the hash after replacing the long name?

I also made some fixes to the docstrings that Ruff complained about. Otherwise looks all good, merging and bumping to `0.9.3`. :)

vd-dragan21:
Thank you for the fast review and release 🤩!
I will add more tests, for now the files you mentioned are being handled differently, since the whole filename including suffix is being replaced by a hash.
Handling suffix is a good idea. I didn't introduce it for 2 reasons:
- in case of API request, there are usually no extension suffix (and in case of File Download, the actual filename is used and not the part of the URL)
- in case of the API request that I'm sending, the suffix (everything after the last "." in the URL) is confusing, e.g. `"*RLRAEAQVK*)&select=structure_id,structure_descriptions,linear_sequence&order=structure_id"` for the following URL:

`
https://query-api.iedb.org/epitope_search?or=(linear_sequence.ilike.*IVLPEDKSW*,linear_sequence.ilike.*ALGIGILTV*,linear_sequence.ilike.*LSLRNPILV*,linear_sequence.ilike.*PKYVKQNTLKLAT*,linear_sequence.ilike.*EIYKRWII*,linear_sequence.ilike.*LLDFVRFMGV*,linear_sequence.ilike.*RLRAEAQVK*)&select=structure_id,structure_descriptions,linear_sequence&order=structure_id
`
But I could set a limit to the suffix length, for example, to avoid too long suffixes that are actually not extensions.

I also saw the Ruff warnings you fixed and was not sure whether I should introduce all the changes to the parts of the files I didn't change. But now I know.

slobentanzer:
@vd-dragan21 agree, there is still room for improvement, but we can do it incrementally (most important to not break anything)

Re the Ruff warnings: I typically fix them in a "drive-by" manner, if I am working on a file anyways. I also don't fix all of them; if the fix is more complex than just restructuring a docstring, it requires some consideration. But if it's an easy fix, why not. :)

slobentanzer:
@vd-dragan21 PS: just be careful when submitting a PR; if you formatting changes are too extensive, it is difficult to assess the actual code changes (because everything shows up red and green). So the way you did it here was actually good, the PR was focussed on the code change. I just used the opportunity to fix some warnings in the same file.

PR #436, opened 2025-04-16
Add AnnData KG in memory representation

vd-dragan21:
Created AnnDataKG Class to be able to generate Scirpy-friendly KG output.
Tests are in progress.


slobentanzer (commented):
Adding a comment about the refactor also here: as previously discussed, would be good to
- rename module and class to better reflect function
- refactor main method into individual methods (retracing the individual steps of the main method); can be private methods (method name led by underscore)
- add individual unit tests for each of those private methods
- migrate function(s) that are used by all child classes to the parent class

Feel free to re-request my review once implemented :)

vd-dragan21:
@slobentanzer I wanted to know your opinion on how to handle the scirpy package installation.
I'm using some functionalities implemented in scirpy for AIRR-AnnData conversion when creating AnnData in-memory representation. That means that the biocypher lock file will include scirpy and everything that belongs to it, which might not be optimal because every Biocypher user will install scirpy as a dependency. I can: 
- either still add scirpy to lock file
- or reimplement the functionality from scirpy in biocypher and add anndata and awkward packages instead (which are definitely smaller than scirpy and implementations are also easy)

slobentanzer:
@vd-dragan21 as discussed, would propose optional dependencies. Check the biochatter toml for examples: https://github.com/biocypher/biochatter/blob/main/pyproject.toml

biocypher/output/in_memory/_airr_to_anndata.py
```diff
@@ -0,0 +1,218 @@
...
+            msg = (
+                "scirpy package is required for AIRR to AnnData conversion. "
+                "Install it with 'poetry add biocypher[scirpy]' or 'poetry add scirpy'."
+            )
+            raise ImportError(
+                msg,
+            )
+
+    def get_kg(self, verbose=True):
+        """Convert directly to AnnData instead of going through DataFrames."""
```
slobentanzer, line 37: "instead of going through DataFrames" is not useful for the reader of this docstring. The user does not know the history of the class, nor should they need to.

Instead, focus on a clear description of the method, and please also add the parameter description.

biocypher/output/in_memory/_airr_to_anndata.py
```diff
@@ -0,0 +1,218 @@
...
+        # Store entities by type directly
+        self.entities_by_type = {}
+
+    def _check_dependencies(self):
+        """Verify that scirpy is available."""
+        if not HAS_SCIRPY:
+            msg = (
+                "scirpy package is required for AIRR to AnnData conversion. "
+                "Install it with 'poetry add biocypher[scirpy]' or 'poetry add scirpy'."
+            )
```
slobentanzer, line 31: The module name refers to AnnData, but the dependency check is for scirpy. This is not aligned, should be clear and one or the other.

biocypher/output/in_memory/_airr_to_anndata.py
```diff
@@ -0,0 +1,218 @@
...
+    def _add_to_entities_by_type(self, lists):
+        """Add entities from lists to the entities_by_type dictionary."""
+        for _type, _entities in lists.items():
+            if _type not in self.entities_by_type:
+                self.entities_by_type[_type] = []
+            self.entities_by_type[_type].extend(_entities)
+
+    def to_airr_cells(self, verbose=False):
+        """Convert BioCypher entities to the AIRR cells."""
+        tra_nodes = self.entities_by_type.get("tra sequence", [])
```
slobentanzer, line 61: I am not sure we should hardcode biocypher labels here, as well as assume that the variables `trb_nodes` etc will cover all use cases. The module has generic claims (see my comment above about whether it should be anndata or scirpy), but this code is very specific to one possible schema in the tcr-epitope project. If the user of the tcr-epitope repo changes only a small detail in the schema, it will break.

Let alone someone wanting to use the anndata creation for an entirely different biocypher project. The modules in biocypher need to be generic, i.e., applicable to all potential use cases in biocypher usage. :)

biocypher/output/in_memory/_airr_to_anndata.py
```diff
@@ -0,0 +1,218 @@
...
+            cell.add_chain(beta_chain)
+        if epitope_ids:
+            add_epitope_metadata(epitope_dict, cell, epitope_ids)
+        cell["data_source"] = "BioCypher"
+        cell["is_paired"] = paired
+        return cell
+
+
+def extract_sequence_from_id(receptor_id):
+    """Extract CDR3 sequence from receptor ID like 'tra:CIRSGSARQLTF'."""
```
slobentanzer, line 196: This method assumes that the formatting of a sequence will always be <something>:<sequence>; is that a fair assumption?

test/fixtures/data_generator.py
```diff
@@ -3,7 +3,7 @@
 from biocypher._create import BioCypherEdge, BioCypherNode, BioCypherRelAsNode
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture()
```
slobentanzer, line 6: Why did you remove the function scope?

pyproject.toml
```diff
@@ -25,6 +25,7 @@ readme = "README.md"
 
 [project]
 requires-python = ">=3.10"
+name = "biocypher"
```
slobentanzer, line 28: This is a repetition of the name parameter above (under tool.poetry)

slobentanzer (changes requested):
Hi @vd-dragan21, thanks, good progress!

Regarding the AIRR to AnnData module, I believe we need to be more generic. We cannot assume that the schema will be defined exactly like that in every project that aims to use the module / class (hardcoding labels such as "trb sequence"); and we cannot rely on the assumption that the hardcoded variables you introduce (such as `trb_nodes`, `epitope_nodes`) will be sufficient to cover all use cases of AnnData creation.

We should find a way to generalise the current approach, after agreeing on the level of generalisation we require. For instance: Is the module targeted at creating AnnData generally or, rather, just the type of AnnData that scirpy expects?

I left some more individual comments about this and other small things.

test/fixtures/data_generator.py
vd-dragan21: I believe it was a ruff suggestion. The motivation behind it is that "function" is already the default scope for pytest fixtures, and it was only specified in one fixture and not the others, so I deleted it for consistency.

slobentanzer: You can see in the commit changes that it was in all 5 fixtures in this module, and it is in many others in the test battery as well. For me, Ruff does not suggest that, are you sure it wasn't AI?

It is good practice to include it because it makes the scope explicit. Please leave it in and add the explicit scope also to the fixtures you introduce.

slobentanzer:
@vd-dragan21 after some thought, I think this module here should be generic, with no assumptions on the data model (biocypher schema). We may want to introduce some quality control and validation for our specific `tar-epitope` use case, however, these should be done there. So, practically, I would hardcode the AIRR labels and everything else that is a convention there in the adapter/pipeline, and include warnings if the user diverges from these conventions (e.g., calling the fields differently). It should however still be possible to represent any schema here in the biocypher module, so I would keep this generic and assume that the `tcr-epitope` repository will implement the checks to make the schema compatible with the AIRR standards.

Sound good, or is there a problem with proceeding like that?

pyproject.toml
vd-dragan21: @slobentanzer which poetry version are you using?
It [seems](https://github.com/aidos-lab/pytorch-topological/issues/43) to be a "new" issue that appears after the introduction of project section support as for PEP 621. This is mentioned in the [release notes](https://github.com/python-poetry/poetry/releases/tag/2.0.0).
I can either downgrade my poetry back to 1.8.5 from 2.1.3 or adjust the pyproject.toml as you can see [here](https://github.com/aidos-lab/pytorch-topological/issues/43) or [here](https://github.com/python-poetry/poetry/commit/a5887d846ab9a8f3c31157ad62d3e923271ee302):
1. Change [tool.poetry] to [project] and move python requirement to the new [project] section
2. Adjust the author's spelling to
```
authors = [
    { name = "Sebastian Lobentanzer", email = "sebastian.lobentanzer@gmail.com" },
    { name = "Denes Turei", email = "turei.denes@gmail.com" },
]
```
instead of
```
authors = [
    "Sebastian Lobentanzer <sebastian.lobentanzer@gmail.com>",
    "Denes Turei <turei.denes@gmail.com>"
]
```

slobentanzer: @vd-dragan21 the issue you mention is known but not related to my comment. I am going back and forth between >2 and <2 because some of our CI is not yet updated to >2. New projects should be >2, but for some legacy projects (like `biocypher`), the change is not trivial.

Now for the problem I initially tagged here: this is not the author name, but the project name, and it is simply in the pyproject twice, which does not cause an error, but is unnecessary.

For the `biocypher` CI, we use Poetry <2 until we have time to update all the workflows to Poetry >2, otherwise the CI will fail.

vd-dragan21: @slobentanzer Sorry, I didn't state it clearly, I'm getting the same error as described [here](https://github.com/aidos-lab/pytorch-topological/issues/43) if I don't add the name to the project section:
```
The Poetry configuration is invalid:
  - project must contain ['name'] properties

```
It appears when using the Poetry version >2.

The new convention for the author names in pyproject.toml was also introduced in the later Poetry versions and should be updated as well, that is why I mentioned it.

If I add the biocypher name to the project section - it is a quick and dirty fix, which makes the pyproject.toml compatible with both, poetry <2 and >2 for me.

slobentanzer: In that case, I would recommend you use locally whichever Poetry the project you're working on uses. Updating Poetry everywhere is not a huge priority TBH; if you want, you can do it for BioCypher (but make sure to evaluate whether the time is worth it).

Reinstalling Poetry to change the version just takes a minute, I typically just do that if I have a project in another Poetry version.

slobentanzer:
@vd-dragan21 did you want me to review this? If so, please make use of the re-request button in the top right of the PR. :)

vd-dragan21:
@slobentanzer I wanted to get the feedback from Gregor first

PR #440, opened 2025-04-30
Fix neo4j import script not correctly generated due to problematic writer instance management

kpto:
When the output is set to be neo4j offline, the generated import script is incorrect. The commands inside will only import one node type. After investigation, this is because a new writer is repeatedly created which makes BioCypher lose track of the CSV files created.

Another bug was found along the way that `write_edges` never creates a writer so that one cannot update an existing KG.

This PR fixes both of them.

I am not sure `main` is the branch I should target but it seems that `dev` is abandoned. In addition, `test_sqlite` does not pass because the script generated in the test is incorrect on Windows platform.

### DIFF
--- a/biocypher/_core.py
+++ b/biocypher/_core.py
@@ -301,7 +301,7 @@ class BioCypher:
 
         return self._translator
 
-    def _get_writer(self):
+    def _initialize_writer(self) -> None:
         """Create writer if not online.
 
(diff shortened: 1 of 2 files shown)


slobentanzer (approved):
Hi @kpto, many thanks for the fix! Merging and bumping to `0.9.4`

Issue #442, opened 2025-05-12
[BUG]: Default bin prefix in neo4j writer


ecarrenolozano:
### Source

poetry

### Python version

3.12

### BioCypher version

biocypher v0.9.3

### OS platform and distribution

Linux Ubuntu 22.04 64-bit

## Issue:
- Once a `neo4j-admin-import-call-sh` has been generated. Only the `version` section is ignoring the parameter defined in the `biocypher_config.yaml` file.

## Input

`biocypher_config.yaml` file:
```yaml
neo4j:
  delimiter: '\t'
  array_delimiter: "|"
  skip_duplicate_nodes: true
  skip_bad_relationships: true
  import_call_bin_prefix: /opt/neo4j-path/  # path to "neo4j-admin"
```

## Output

![Image](https://github.com/user-attachments/assets/43ccc4a3-06a4-42cc-9081-d436bd182512)

### Relevant LOG output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mbaric758:
Hello, 

Would it be possible to adjust so that the import bin prefix is also updated in the `version` call, based on what is specified in the `import_call_bin_prefix:` parameter? Since it should match how `neo4j-admin` is called in the `if-else` loops below. Otherwise there's errors in case of using `neo4j-admin` as a global variable.

mbaric758:
Yes, for example, if we put `import_call_bin_prefix: ""` it should also affect the prefix in the `version` section of the call, before the `if-else` loop. What I currently get as output with this setting is:

<img width="441" alt="Image" src="https://github.com/user-attachments/assets/051336ad-ceb0-4dcf-8e12-14efd3eecec2" />

So here the default prefix is still present in the version detection section. Ideally, I think, it should match what is defined in the `import_call_bin_prefix`. So in this example above, the output should be:

<img width="409" alt="Image" src="https://github.com/user-attachments/assets/cd957985-d00e-4b10-a0d7-69ee5a6be158" />

What do you think?

mbaric758:
Ah, I see - it was a problem on my side, I was on an older version. Thank you!

PR #443, opened 2025-05-12
Fix default import call bin prefix in neo4j writer.

ecarrenolozano:
This Pull Request fixes the issue #442 

Files modified:
- `biocypher/output/write/graph/_neo4j.py`

### Logic behind the PR
The logic is simply to reuse the same argument—`self.import_call_bin_prefix`—which is already used to update the content of the if-else statement. This attribute is inherited from the abstract class that defines the batch writer. All necessary checks for cases where this parameter is not defined in the configuration file are handled within the abstract class. The implementation works correctly even when the parameter is set to None or commented out (e.g., `#import_call_bin_prefix: /opt/neo4j-path`), as the _batch_writer includes logic to handle such cases. 
```python
## Code in _batch_writer.py
if import_call_bin_prefix is None:
    self.import_call_bin_prefix = self._get_default_import_call_bin_prefix()
else:
    self.import_call_bin_prefix = import_call_bin_prefix
```


slobentanzer:
Hi @ecarrenolozano,
Thanks for the fix! I don't fully get the reason behind the change, would be great to have some explanation about the PR.

Can you describe how the PR addresses the linked issue? What has changed about the logic?

I don't see a check for the assignment of `self.import_call_bin_prefix`; what happens if this is not defined in a pipeline's config; can it be `None`? What if the user wants to use the default (`bin/neo4j`)? The previous method (the line you commented out) provided a hardcoded default, which ensured that there is at least always a value there. How does this change in your solution?

If we don't require `_get_default_import_call_bin_prefix()` any more, we should delete this method. It is also better practice to remove unneeded lines of code, not comment them out.

ecarrenolozano:
Hi @slobentanzer, in 

> Thanks for the fix! I don't fully get the reason behind the change, would be great to have some explanation about the PR.

If I define the path to the Neo4j binaries in the configuration file (e.g., `import_call_bin_prefix: /opt/neo4j-path`), the change only takes effect within the if-else statement section. The version section does not reflect this update and always uses the default binary directory. Just let me know if there's any part that's still unclear—happy to clarify!

> 

> Can you describe how the PR addresses the linked issue? What has changed about the logic?
> 
> I don't see a check for the assignment of self.import_call_bin_prefix; what happens if this is not defined in a pipeline's config; can it be None? What if the user wants to use the default (bin/neo4j)? The previous method (the line you commented out) provided a hardcoded default, which ensured that there is at least always a value there. How does this change in your solution?
> 
> If we don't require _get_default_import_call_bin_prefix() any more, we should delete this method. It is also better practice to remove unneeded lines of code, not comment them out.

Certainly. The logic is simply to reuse the same argument—`self.import_call_bin_prefix`—which is already used to update the content of the if-else statement. This attribute is inherited from the abstract class that defines the `batch writer`. All necessary checks for cases where this parameter is not defined in the configuration file are handled within the abstract class. The implementation works correctly even when the parameter is set to None or commented out (e.g., #import_call_bin_prefix: /opt/neo4j-path), as the _batch_writer includes logic to handle such cases. In this case it calls the default path. As I said, all the checks have been already implemented in the abstract class

```python
if import_call_bin_prefix is None:
    self.import_call_bin_prefix = self._get_default_import_call_bin_prefix()
else:
    self.import_call_bin_prefix = import_call_bin_prefix
```
Finally, the _get_default_import_call_bin_prefix(self) inside the _neo_4j.py overrides the behavior of the same function in _batch_writer.py. So, it is not a good idea to remove this function. The reason I comment the solution is just in case of an extra explanation like in this case. I will remove it.

PR #449, opened 2025-06-10
Isolate graph class test

vd-dragan21:
Added example of the refactored isolated Graph class, added method to transform to adj_list.

### DIFF
--- a/biocypher/_core.py
+++ b/biocypher/_core.py
@@ -172,10 +172,6 @@ class BioCypher:
         self._driver = None
         self._in_memory_kg = None
 
-        self._in_memory_kg = None
-        self._nodes = None
-        self._edges = None
-
     def _initialize_in_memory_kg(self) -> None:
         """Create in-memory KG instance.
 
@@ -224,10 +220,6 @@ class BioCypher:
 
         TODO: to_df implies data frame, should be specifically that use case
         """
-        return self._to_KG()
-
-    def to_networkx(self):
-        """Create networkx using internal representation."""
         return self._to_KG()
 
     def _to_KG(self):
@@ -384,9 +376,10 @@ class BioCypher:
         g = Graph()
         g.add_nodes(nodes)
 
-        if not self._translator:
-            self._get_translator()
-        translated_nodes = self._translator.translate_entities(g.nodes)
+        # if not self._translator:
+        #     self._get_translator()
+        # translated_nodes = self._translator.translate_entities(g.nodes)
+        translated_nodes = g.nodes
 
         if self._offline:
             if not self._writer:
@@ -553,7 +546,7 @@ class BioCypher:
         if not self._in_memory_kg:
             msg = "No in-memory KG instance found. Please call `add()` first."
             raise ValueError(msg)
-
+            
         if not self._in_memory_kg:
             self._initialize_in_memory_kg()
         return self._in_memory_kg.get_kg()
--- a/biocypher/output/in_memory/_get_in_memory_kg.py
+++ b/biocypher/output/in_memory/_get_in_memory_kg.py
@@ -8,24 +8,25 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from biocypher._logger import logger
-from biocypher.output.in_memory._networkx import NetworkxKG
-from biocypher.output.in_memory._pandas import PandasKG
+# from biocypher.output.in_memory._networkx import NetworkxKG
+# from biocypher.output.in_memory._pandas import PandasKG
+from biocypher.output.in_memory._graph import Graph
 
 if TYPE_CHECKING:
     from biocypher._deduplicate import Deduplicator
-    from biocypher.output.in_memory._in_memory_kg import _InMemoryKG
+    # from biocypher.output.in_memory._in_memory_kg import _InMemoryKG
 
 logger.debug(f"Loading module {__name__}.")
 
 __all__ = ["get_in_memory_kg"]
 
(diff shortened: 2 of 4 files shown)


biocypher/output/in_memory/_graph.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def __init__(self, deduplicator=None, output_format="adj_list"):
+        """Initialize an empty graph with no nodes or edges."""
+        self.nodes: List[NodeInfo] = []
+        self.edges: List[EdgeInfo] = []
+        self._node_ids: Dict[str, int] = {}
+        self.deduplicator = deduplicator
+        self.output_format = output_format
+
+    def get_kg(self):
```
kpto, line 60: I would recommend having an abstract class `BaseGraph` and the derived classes `NetworkXBackedGraph`, `PandasBackedGraph` and `AdjacentListBackedGraph`. You don't need separate conversion methods, just override the abstract method for each derived class.

kpto: This also avoids the need of magic strings like `networkx` and `adj_list`. But if such strings cannot be avoided, I highly recommend to use enum instead, slightly more effort at first but makes refactoring later a lot easier.

biocypher/output/in_memory/_graph.py
```diff
@@ -0,0 +1,247 @@
...
+            node_id: The ID of the node to retrieve.
+
+        Returns:
+            The NodeInfo object if found, None otherwise.
+        """
+        if node_id in self._node_ids:
+            return self.nodes[self._node_ids[node_id]]
+        return None
+
+    def get_node_edges(self, node_id: str) -> List[EdgeInfo]:
```
kpto, line 131: Do you mean `get_edges`?

biocypher/output/in_memory/_graph.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+    An in-memory graph data structure to store nodes and edges.
+
+    This class provides functionality to create and manage a graph structure
+    compatible with BioCypher's data model.
+    """
+
+    def __init__(self, deduplicator=None, output_format="adj_list"):
+        """Initialize an empty graph with no nodes or edges."""
+        self.nodes: List[NodeInfo] = []
```
kpto, line 54: Hide the field with an underscore. If it needs to be exposed, expose it as an `Iterable` or `Sequence` property so the user cannot modify it accidentally.

Same for `self.edges`

biocypher/output/in_memory/_graph.py
```diff
@@ -0,0 +1,247 @@
...
+        self.edges: List[EdgeInfo] = []
+        self._node_ids: Dict[str, int] = {}
```
kpto, line 56: I usually name these things in this way: `node_id_index_map`

biocypher/output/in_memory/_graph.py
```diff
@@ -0,0 +1,247 @@
...
+        self.deduplicator = deduplicator
```
kpto, line 57: Should be hidden?

biocypher/output/in_memory/_graph.py
```diff
@@ -0,0 +1,247 @@
...
+        self.output_format = output_format
```
kpto, line 58: Should be hidden?

biocypher/output/in_memory/_graph.py
```diff
@@ -0,0 +1,247 @@
...
+
+class Graph:
+    """
+    An in-memory graph data structure to store nodes and edges.
+
+    This class provides functionality to create and manage a graph structure
+    compatible with BioCypher's data model.
+    """
+
+    def __init__(self, deduplicator=None, output_format="adj_list"):
```
kpto, line 52: Missing return type? Same for all other methods missing return type.

biocypher/output/in_memory/_graph.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        return [edge for edge in self.edges if edge.source_id == node_id or edge.target_id == node_id]
+
+    def clear(self) -> None:
+        """Remove all nodes and edges from the graph."""
+        self.nodes = []
+        self.edges = []
+        self._node_ids = {}
+
+    def to_adj_list(self) -> Dict[str, List[str]]:
```
kpto, line 149: Can't you just type the values as `Tuple[List[str], Dict[str, Any]]` so it becomes a properties carrying adjacent list?

biocypher/_core.py
```diff
@@ -378,9 +372,14 @@ def _add_nodes(
             `IN_MEMORY_DBMS`
 
         """
-        if not self._translator:
-            self._get_translator()
-        translated_nodes = self._translator.translate_entities(nodes)
+        # Create a list inside of Graph
+        g = Graph()
```
kpto, line 376: Confused here, why can you instantiate a `Graph` without arguments?

biocypher/output/in_memory/_get_in_memory_kg.py
```diff
@@ -8,24 +8,25 @@
...
 
-IN_MEMORY_DBMS = ["csv", "pandas", "tabular", "networkx"]
+IN_MEMORY_DBMS = ["csv", "pandas", "tabular", "networkx", "adj_list"]
 
 
 def get_in_memory_kg(
     dbms: str,
     deduplicator: Deduplicator,
-) -> _InMemoryKG:
+):
```
kpto, line 29: Why was the return type removed?

PR #458, opened 2025-07-25
perform type checking to prevent import failure

slobentanzer:


slobentanzer:
@jdreo thanks :)

@vd-dragan21 as soon as you're ready, please approve so we can merge.

PR #472, opened 2025-12-05
Update _networkx.py

hdraisma:
REF "https://stackoverflow.com/a/47197029", and as at runtime I was getting the below ModuleNotFoundError:
```
    G_loaded = pickle.load(f)
ModuleNotFoundError: No module named 'networkx'
```


slobentanzer:
closing as seems stale; feel free to open if I'm mistaken

PR #501, opened 2026-04-15
refactor(batch_writer): applies property types from the schema

jdreo:
Previous implementation was casting down everything as string, without any check, and with duplicated code for nodes and edges.

This implementation casts the input property values to the type configured in the schema, before serializing to string, thus avoiding format issues downstream.


jdreo:
I don't have time in the coming weeks to setup a complete PR with tests, doc and config management. Help welcomed :-)

slobentanzer:
I can see if I can find some time to review / fix.

PR #502, opened 2026-04-20
Remove unused Python shebangs from example scripts

### LINKED ISSUE #383: Remove Python shebang and copyright comments from the top of modules

mvanhorn:
Closes #383.

The two example scripts still carried a `#!/usr/bin/env python3` line
even though they are not marked executable and the docs invoke them as
`python examples/<name>.py`. The shebangs are dead code, so drop them
per the direction in #383.

Sweep confirmation:

```bash
$ rg '^#!/' --glob '*.py' --glob '*.pyi'
examples/unified_graph_example.py:1:#!/usr/bin/env python3
examples/agent_integration_example.py:1:#!/usr/bin/env python3
```

After this PR, no Python file in the repo starts with a shebang. The
companion request in #383 ("copyright comments from the top of modules")
already appears to be clean - I could not find any author/copyright
headers in any `*.py` file.

### Verification

Both files still parse cleanly after the edit (docstring is now the
module's first statement).

```bash
$ python -c 'import ast; ast.parse(open("examples/unified_graph_example.py").read()); ast.parse(open("examples/agent_integration_example.py").read())'
```

### DIFF
--- a/examples/agent_integration_example.py
+++ b/examples/agent_integration_example.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 """BioCypher Agent Integration Example
 
 This example demonstrates how LLM agents can use BioCypher to create and
--- a/examples/unified_graph_example.py
+++ b/examples/unified_graph_example.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 """Unified Graph Example - Simplified Knowledge Representation
 
 This example demonstrates how the unified Graph class streamlines knowledge


slobentanzer:
Thanks @mvanhorn (or your copilot); the issue was targeted at *module* shebangs; the example scripts are meant to be executed, so the shebangs have a reason there. They also do not include the reason for the issue, which is the license / copyright text that was hard to maintain. In any case, the issue is erroneously open still, we removed these shebangs a while ago; I don't remember, but the modules seem to not contain any. Closing this PR and the issue.

PR #510, opened 2026-05-10
fix(batch_writer): write boolean properties as lowercase true/false for Neo4j

### LINKED ISSUE #368: Boolean sometimes not imported correctly (all 'false')
Need to check the implementation for writing booleans to Neo4j admin import files; happened that regular Python bool values were all imported as 'false' (Neo4j seems to write bools lower case).

slobentanzer:
## Summary

Fixes #368

- Python's `str(True)` produces `'True'` (capital T), which Neo4j admin import does not recognise as a boolean value, causing all boolean properties to be silently imported as `false`.
- Fix: use `str(p).lower()` for `bool`/`boolean` property types so the CSV output contains lowercase `'true'`/`'false'` as required by Neo4j.
- The fix applies in both `_write_node_data` and `_write_edge_data` in `_batch_writer.py` (and consequently `_ArangoDBBatchWriter` which inherits from the Neo4j writer).

## Changes

- `biocypher/output/write/_batch_writer.py`: split the numeric/boolean type branch so `bool`/`boolean` values are serialised with `.lower()`.
- `test/output/write/graph/test_neo4j.py`:
  - Update existing `test_BioCypherRelAsNode_implementation` assertion from `"True"` to `"true"` to match correct output.
  - Add `test_write_edge_data_boolean_properties` which explicitly creates edges with `True`/`False` boolean properties and asserts that the CSV contains lowercase `"true"`/`"false"` and does **not** contain Python-style `"True"`/`"False"`.

## Test plan

- [x] `test_BioCypherRelAsNode_implementation` passes (asserts `true` in node CSV)
- [x] `test_write_edge_data_boolean_properties` passes (new test)
- [x] All 39 existing Neo4j writer tests pass
- [x] ArangoDB writer tests pass

https://claude.ai/code/session_01Qo13wdZz9PfFR6i9DHy9u9

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qo13wdZz9PfFR6i9DHy9u9)_

### DIFF
--- a/biocypher/output/write/_batch_writer.py
+++ b/biocypher/output/write/_batch_writer.py
@@ -754,6 +754,8 @@ class _BatchWriter(_Writer, ABC):
                     p = n_props.get(k)
                     if p is None:  # TODO make field empty instead of ""?
                         plist.append("")
+                    elif v in ["bool", "boolean"]:
+                        plist.append(str(p).lower())
                     elif v in [
                         "int",
                         "integer",
@@ -761,8 +763,6 @@ class _BatchWriter(_Writer, ABC):
                         "float",
                         "double",
                         "dbl",
-                        "bool",
-                        "boolean",
                     ]:
                         plist.append(str(p))
                     elif isinstance(p, list):
@@ -991,6 +991,8 @@ class _BatchWriter(_Writer, ABC):
                 p = e_props.get(k)
                 if p is None:  # TODO make field empty instead of ""?
                     plist.append("")
+                elif v in ["bool", "boolean"]:
+                    plist.append(str(p).lower())
                 elif v in [
                     "int",
                     "integer",
@@ -998,8 +1000,6 @@ class _BatchWriter(_Writer, ABC):
                     "float",
                     "double",
                     "dbl",
-                    "bool",
-                    "boolean",
                 ]:
                     plist.append(str(p))
                 elif isinstance(p, list):
--- a/test/output/write/graph/test_neo4j.py
+++ b/test/output/write/graph/test_neo4j.py
@@ -801,6 +801,40 @@ def test_write_edge_data_from_list_no_props(bw):
     assert "\n" in is_mutated_in
 
 
+def test_write_edge_data_boolean_properties(bw):
+    """Boolean properties must be written as lowercase 'true'/'false' for Neo4j admin import."""
+    edges = [
+        BioCypherEdge(
+            relationship_id="gg1",
+            source_id="g1",
+            target_id="g2",
+            relationship_label="gene to gene association",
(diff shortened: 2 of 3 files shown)


slobentanzer:
@ecarrenolozano this is an automated Claude PR (was told to pick an issue and solve it). I am wondering whether this was still a problem, or maybe the issue was outdated. Can you check please?

biocypher/output/write/_batch_writer.py
```diff
@@ -754,15 +754,15 @@ def _write_single_node_list_to_file(
                     p = n_props.get(k)
                     if p is None:  # TODO make field empty instead of ""?
                         plist.append("")
+                    elif v in ["bool", "boolean"]:
```
ecarrenolozano, line 757: Great, it covers booleans in node properties.

biocypher/output/write/_batch_writer.py
```diff
@@ -991,15 +991,15 @@ def _write_single_edge_list_to_file(
                 p = e_props.get(k)
                 if p is None:  # TODO make field empty instead of ""?
                     plist.append("")
+                elif v in ["bool", "boolean"]:
```
ecarrenolozano, line 994: Great, it cover booleans in edge properties.

test/output/write/graph/test_neo4j.py
```diff
@@ -801,6 +801,40 @@ def test_write_edge_data_from_list_no_props(bw):
     assert "\n" in is_mutated_in
 
 
+def test_write_edge_data_boolean_properties(bw):
```
ecarrenolozano, line 836: Missing: Could we also add a direct node boolean-property test?

uv.lock
```diff
@@ -277,7 +277,7 @@ wheels = [
 
 [[package]]
 name = "biocypher"
-version = "0.13.5"
+version = "0.13.6"
```
ecarrenolozano, line 280: Looks good to me.

ecarrenolozano (commented):
The implementation looks correct to me, and I verified the Neo4j writer tests locally with:

`uv run --extra neo4j pytest test/output/write/graph/test_neo4j.py -q`

However, I think we should add a direct node boolean-property regression test before merging. This PR changes both node and edge serialization paths, but the new explicit test only covers edge properties.

slobentanzer:
@ecarrenolozano thanks, do you just want to add the missing test?

ecarrenolozano:
> @ecarrenolozano thanks, do you just want to add the missing test?

Yes, I would like to.

ecarrenolozano (approved):

ecarrenolozano (approved):

slobentanzer:
@ecarrenolozano thanks!

PR #512, opened 2026-05-11
fix(neo4j): "escape" colons in property names

jdreo:
Neo4j's docs says property names can be escaped with backticks: https://neo4j.com/docs/cypher-manual/current/syntax/naming/
However, this does not take precedence over the colon, which
is used to indicate the property type.

Thus, escaping a string with back quotes hoping that
it captures the colon won't work.
Similarly, double-quoting or \: don't work either.
And we are left only with replacing colon with a `_COLON_`  placeholder.

### DIFF
--- a/biocypher/output/write/graph/_neo4j.py
+++ b/biocypher/output/write/graph/_neo4j.py
@@ -64,8 +64,20 @@ class _Neo4jBatchWriter(_BatchWriter):
         return "bin/"
 
     def _quote_string(self, value: str) -> str:
-        """Quote a string. Quote character is escaped by doubling it."""
+        """Quote a value string. Quote character is escaped by doubling it."""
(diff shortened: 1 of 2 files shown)


slobentanzer:
@jdreo thanks! Is there an issue describing the problem? I am trying to understand motivations. What would be a case where we rely on colons being in the property name?

An alternative I can think of is to provide the user with a choice, e.g., "remove colons" or "replace colons with `x`" (in the config). Might be less complex than using `_COLON_` placeholders and then have to convert back in a second step in the DB (which, if Neo4j, might not even be possible either, because it messes with the query language).

I think, within the confines of the Neo4j output module, we should assume the users want something that aligns with Neo4j practice (which would not use colons as far as I can see). How can we help the user do that?

Maybe an issue with 2 use cases would help me understand, if it doesn't already exist. :)

PR #515, opened 2026-05-15
fix(config): properly merge all three config levels in read_config()

slobentanzer:
## Summary

Fixes a bug in `read_config()` where user-level config settings were silently dropped whenever the same top-level key also appeared in the local project config.

**The docstring says:**
> "Read and merge the built-in default, the user level and directory level configuration, with the later taking precedence over the former."

**What actually happened (before this fix):**

```python
# Old logic — picks ONE source per top-level key
value = local[key] if key in local else user[key] if key in user else None
```

If a user had `neo4j.uri` in `~/.config/biocypher/conf.yaml` AND the project had a local `biocypher_config.yaml` with only `neo4j.database_name`, the user's `uri` was silently lost because `local` "won" the whole `neo4j` section.

**What happens now:**

All three levels are applied in precedence order (defaults → user → local), so each setting is overridden only by a more specific config level:

```python
# Apply user settings on top of defaults first
if key in user and user[key] is not None:
    defaults[key].update(user[key])   # or replace for scalar values

# Apply local settings on top (local wins)
if key in local and local[key] is not None:
    defaults[key].update(local[key])
```

## Changes

- `biocypher/_config/__init__.py`: replace the single-source selection with two sequential merge steps
- `test/test_config.py`: add `test_read_config_merges_user_and_local` — asserts that settings from user config, local config, and defaults are all present in the result

## Test plan

- [x] `test_read_config_merges_user_and_local` — new test confirming the three-way merge works correctly
- [x] `test_read_yaml` and `test_for_special_characters` continue to pass
- [x] `test_core.py`, `test_translate.py`, `test_mapping.py` — 36 tests, all pass

---
_Generated by [Claude Code](https://claude.ai/code/session_01ED6bEPZmitZTANpkftQsZj)_

### DIFF
--- a/biocypher/_config/__init__.py
+++ b/biocypher/_config/__init__.py
@@ -82,13 +82,19 @@ def read_config() -> dict:
     local = _read_yaml("biocypher_config.yaml") or _read_yaml("config/biocypher_config.yaml") or {}
 
     for key in defaults:
-        value = local[key] if key in local else user[key] if key in user else None
-
-        if value is not None:
-            if isinstance(defaults[key], str):  # first level config (like title)
-                defaults[key] = value
+        # Apply user-level settings on top of defaults first
+        if key in user and user[key] is not None:
+            if isinstance(defaults[key], dict):
+                defaults[key].update(user[key])
             else:
-                defaults[key].update(value)
+                defaults[key] = user[key]
+
+        # Apply local settings on top (local takes precedence over user)
+        if key in local and local[key] is not None:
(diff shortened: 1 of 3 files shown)


ecarrenolozano (approved):
This PR passes all tests and fixes the problem. In case you envisioned more complex scenarions between config files, probably it is time to consider to migrate to Omegaconf.

slobentanzer:
@ecarrenolozano thanks!

PR #522, opened 2026-05-21
feat(neo4j): parquet format output for batch import

kpto:
Closes https://github.com/biocypher/biocypher/issues/492

### Problem

CSV is the default and most commonly used format for Neo4j batch import, but it has several drawbacks:

1. Stringified values significantly increase memory usage, creating storage overhead for output files and additional memory pressure during BioCypher runs.
2. CSV is highly susceptible to character conflicts. The current implementation relies on three special characters: the string quote, the CSV column delimiter, and the array delimiter. Combined, these substantially increase the likelihood of conflicts with user data, which has already proven problematic.

### Improvement

Neo4j supports Parquet as an input format while still using the existing CSV header files. Since output data can remain in native data types, memory consumption is reduced and runtime performance improves because value stringification is avoided.

A small benchmark showed that Parquet output with Zstandard compression is approximately one quarter the size of the equivalent CSV output.

Parquet also stores columns and arrays natively, eliminating the character conflict issues inherent to CSV.

### Specification

1. The existing `file_format` output parameter is used:

   * `csv` remains the default
   * `parquet` enables this feature
   * no other values are accepted
2. Output filenames remain unchanged except for the extension:

   * `{Group}-part{part_num}.csv`
   * becomes `{Group}-part{part_num}.parquet`
3. Import script generation remains unchanged, as Neo4j detects the data file format automatically.
4. CSV header generation required a small adjustment due to an apparent Neo4j issue:

   * when arrays are stored as native Parquet arrays rather than delimiter-separated strings, declaring a type such as `int[]` in the header causes import errors
   * declaring the field as `int` works correctly for both scalar and array values during import

### Other Fixes

1. In `_batch_writer.py` (lines 628 and 885), list type inference incorrectly returned `list` instead of `{element_type}[]`, causing array data types to be omitted from the header file.
2. In `_batch_writer.py` (lines 781 and 1020), `_write_array_string` assumed all list elements were strings, causing failures for non-string arrays. This was fixed by converting all elements to strings before writing.

### Remarks

After inspecting the Neo4j source code, the Parquet implementation appears to be largely adapted from the CSV import pipeline. Values are still converted to strings internally rather than being used as native types, which limits some of the potential benefits.

Additionally, the special `:LABEL` field does not accept array values. However, because labels already impose strict character constraints, Parquet still provides meaningful value by avoiding delimiter and escaping conflicts elsewhere.

Writing Parquet files is also more efficient than writing plain text files.

Only `/test/output/write` has been executed due to the scale of the full test suite. No dedicated tests were added because the current testing infrastructure is tightly coupled to CSV output and extending it to support content validation for Parquet would require substantial changes.

Given the current state of Neo4j's Parquet support, I would also suggest treating this as an experimental feature for now rather than officially supporting it.

Feedback on testing strategy and direct changes to the branch are welcome.


kpto:
Strangely, `# noqa: E712` has to be placed on line 1048 in `_batch_writer.py` for the tests to pass, but my formatter (`ruff` as well) insists on moving it to line 1045.

slobentanzer (changes requested):
Hi @kpto, thanks for the submission. Re feedback on testing strategy: I couldn't see any. Agree that we need some; in the regular (CSV) batch writer case, we go by the Neo4j definitions to check output formats to be consistent with what Neo4j expects. I don't see a way around this, so we'd need to do the same for the Parquet format Neo4j expects. If Neo4j are sensible, this should not change inside a major version. We could use the Docker image we are already using in the CI for some procedural or smoke tests.

As far as I can tell, the `map(str, …)` non-string-array fix and the `list→{elem}[]` type-inference fix are correct and also improve the CSV path.

> Only /test/output/write has been executed due to the scale of the full test suite.

By "executed", you mean you didn't run the full suite locally? CI needs to pass on the full suite in any case...

> No dedicated tests were added because the current testing infrastructure is tightly coupled to CSV output

Not sure I understand – what is coupled? The current Neo4j writer tests / fixtures are CSV-directed, yes; but that is not an argument to not have tests for Parquet functionality. You don't need to touch the CSV tests (or any of the other ones we have in the write module). The testing infrastructure is just pytest. Why not just add the tests for Parquet to ensure basic functionality, in a new module?

My observations on TODOs for the PR:
- [ ] Add Parquet tests to the suite. Currently no test exercises `file_format="parquet"`. Tests should cover: write→read round-trip, header/data column alignment, array properties, None/empty-list/mixed-type columns, and the .parquet extension + zstd output.
- [ ] Verify the Parquet import end-to-end, and reconcile the header/import-call story. The PR states the import script is intentionally unchanged because Neo4j auto-detects the format; I haven't fully grasped whether it still relies on the header though. `_get_import_call` (_neo4j.py:434-476) and `_get_import_call_windows` (:478-508) contain no reference to `file_format` and still pass `--delimiter`/`--array-delimiter`/`--quote`, and the header is still emitted as .csv. Add a test (or CI integration test) proving `neo4j-admin … import` actually ingests the generated Parquet output. We do already use a Neo4j image in CI, so should be feasible without bloating CI; can switch to a newer v5 one if the current versions predate Parquet import support. If there's a bug or quirk in Neo4j, we should be able to confirm / test that reproducibly. Tests should align with Parquet-specific logic (`strip("[]")`); either confirm Neo4j consumes the .csv header for Parquet (and test it), or drop the dead header/strip logic.
- [ ] Make pyarrow an optional dependency. Currently, `pyarrow>=24.0.0` would be a hard requirement for all users (and it's not a slim dep). Move it to an optional extra (e.g. biocypher[parquet]) with a lazy import + clear error if missing. Also the (quite recent) `>=24` floor might be unnecessarily high; we could lower the floor to the minimum version the API needs.
- [ ] Document the feature for users. The `file_format="parquet"` option is undocumented. We should add usage docs covering how to enable it, the resulting file layout, and the import workflow/limitations; with a link to Neo4j docs of the feature. We could also mention it in the tutorials, some of which refer to Neo4j. @ecarrenolozano has improved them last.
- [ ] Minor: Harden type handling. Use `.removesuffix("[]")` instead of `.strip("[]")` (strips chars, not a suffix); decide intended behaviour for empty-list properties (currently left untyped); replace the unreachable bare `raise AssertionError` in `_write_array_string` with a `ValueError` or remove it.

kpto:
@slobentanzer Thanks for the input. One of the reasons for opening this PR is to let GitHub run the full test suite for me, as it took quite long to run locally on my laptop.

Regarding the tests, I opened this PR first to get your opinion before implementing them because, with your proposed approach, essentially all content verification tests would be duplicated. I personally dislike this approach because it relies on future maintainers remembering to update the parquet-specific cases whenever changes are made, and outdated tests will not necessarily fail. I think this increases the maintenance burden over time. Ideally, the existing tests should be refactored to be parameterised and select the corresponding content validator based on the format parameter. That way, the responsibility is made explicit in the tests themselves, making future omissions less likely. That said, if you are comfortable with the duplication approach, I can implement it that way as well.

The CSV headers are still required. As mentioned earlier, Neo4j parquet support is essentially an adaptation of the CSV logic, so apart from the underlying format being parquet instead of CSV, there are not many behavioural differences.

Another reason for opening this PR was to clarify whether you want to treat this feature as officially supported. Since you requested documentation updates, I assume the answer is yes.

Lastly, `AssertionError` signals that I believe this code path should never be reached, whereas `ValueError` signals that I expect this situation to occur as part of normal execution. They communicate different intentions. Removing the assertion would result in returning None, which also does not match my intent. The ideal solution would be `typing.assert_never()`, but that requires Python 3.11 and a well typed codebase. Apart from this point, I am fine with all the other requests.

slobentanzer:
@kpto thanks for the explanation.

> One of the reasons for opening this PR is to let GitHub run the full test suite for me, as it took quite long to run locally on my laptop.

Sure, no problem. I just wondered why you stated it.

> all content verification tests would be duplicated. I personally dislike this approach

I dislike it as well, of course; I took your initial comment in the PR message to mean that you don't want to do it sustainably (because of effort). I would be in favour of keeping everything as maintainable as possible. If it is within scope to fix now, we don't need to duplicate. Were you just asking whether it would be ok to make "substantial changes"?

> The CSV headers are still required. As mentioned earlier, Neo4j parquet support is essentially an adaptation of the CSV logic, so apart from the underlying format being parquet instead of CSV, there are not many behavioural differences.

👍

> Another reason for opening this PR was to clarify whether you want to treat this feature as officially supported. Since you requested documentation updates, I assume the answer is yes.

IO is the biggest bottleneck in the KG process; we don't need to take responsibility for the Neo4j side (*iff* we have suitable tests to verify biocypher behaviour), but having a more effective ingestion pathway would definitely be useful for the users. Better to make sure they know it's there (we can say "experimental" when we point to Neo4j documentation state).

> AssertionError signals that I believe this code path should never be reached

I did not know that convention. Understand the sentiment, but for maintainability I would add an explanation. Also looked at explanations online, most sources call it an antipattern (if raised manually). In absence of solid typing, Claude recommended `RuntimeError(f"unhandled file_format: {self.file_format!r}")`; that at least would explain itself (and wouldn't carry the semantic inaccuracy of "assertion"). Does that work?

I tried to introduce typing after the fact at one point; it was too much work. Maybe one of these days an AI copilot can do it.

kpto:
It’s not wrong that I didn’t want to do it, because it would imply a commitment to ongoing maintenance, user expectations, and confirmation of architectural changes, all of which are business decisions with real weight.

I also don’t know the scale of data BioCypher intends to target. My use case — around a billion records — is probably an outlier, and BioCypher may not want to optimise or canonicalise around that scenario. In that case, CSV may already be sufficient, while Parquet could remain an experimental “nice to have” feature without long-term commitment. That approach avoids unnecessary maintenance burden and preserves flexibility. Some features are better positioned that way.

Since you want to make the feature official, tests are certainly warranted. That said, features that are rarely used do not become battle-tested in practice, and unit tests cannot fully compensate for that. If Parquet support becomes an official feature, I would even suggest making it the default format later so it receives broader real-world usage.

Lastly, to me, unreachable code should fail with something intentionally non-business-semantic. A meaningful domain-specific error can mislead future maintainers into thinking there is a valid business-logic path that triggered it, whereas a deliberately generic failure clearly communicates: “this should never happen.” That conveys the implementer’s intent more accurately and helps narrow the problem down to a coding issue rather than a business-logic issue. Even if `RuntimeError` was to be used, I would still prefer a message such as "unreachable code". Personally, I still prefer `AssertionError` because it emphasises the unexpected nature of the state, though `RuntimeError` is acceptable as well.

slobentanzer:
TBH I don't really worry too much about the business decisions; seems like compatibility with legacy is still given. I would handle it like this:
- add the functionality
- add tests (any tests are better than no tests)
- document it
- recommend it for large graphs (or even just for getting around this delimiter / quote character stringification issue)
- make it optional and ensure that the lazy import throws the right error when someone uses parquet in the config
- ensure consistency with legacy processing (which should be simple in this case)

I am not convinced having it as a "hidden" feature is worth doing. Hesitant on making it "default" (whatever that means) because it is an additional large dep and might not be necessary (e.g. due to small size or non-use of Neo4j).

On the `AssertionError`: I didn't understand what it was meant to communicate because I am not familiar with the convention / antipattern. It would probably have been fine with the addition of `(f"unhandled file_format: {self.file_format!r}")`, but since we are talking about it, we can also just make it a `RuntimeError(f"unhandled file_format: {self.file_format!r}")`. Because strictly we are not asserting anything there at all. `AssertionError` also has the disadvantage that it will silently fail if someone runs `python -O`, as assertions are removed in optimised mode.

kpto:
@slobentanzer 
### Changes

1. All existing Neo4j writer tests now cover both CSV and Parquet formats.
2. The dependency has been moved to the optional `bigdata` group.
3. `pyarrow` is now imported lazily.
4. A clear error message is shown when `pyarrow` is required but not installed.
5. Replaced `AssertionError` with `RuntimeError` and added an "unreachable code" message. I still believe this best communicates the intent to future maintainers.

### Remarks

Originally, I was considering positioning this feature as a solution for both character-escaping issues and performance optimisation for large datasets, which is why I suggested making it the default option earlier. Now that the dependency has been moved to the optional `bigdata` group, the character-escaping benefit becomes less visible, even though I consider it the more important advantage.

I'm highlighting this to confirm that this aligns with your intended direction.

I will work on the documentation once this is approved.

By the way, `python -O` removes `assert` statements, but it does not remove the `AssertionError` class.

slobentanzer:
@kpto thanks, agree fully and looks good.

About the string sanitisation, I agree that this is a valuable feature in itself. Is your comment only about the name of the optional dependency group? How would you prefer to position it?

I was of the impression that we'd have other dependencies that relate to big data issues in the dependency group. From another PR? Is that not the case?

Also, we can probably have multiple optional groups that contain the same dependencies (or subsets of them), if it's just about messaging / framing.

I don't disagree with anything you've said, so feel free to go ahead with what you think is the best way forward.

PS: you're right about the AssertionError; it seems that manually raising it is not recommended though (should only occur by `assert` failing, which is removed by `-O`).

kpto:
@slobentanzer I’d personally prefer positioning it as the default output format for Neo4j batch writing, with the required dependency included as part of the core package. I don’t consider 25 MB to be a large package by today’s standards, but if we do want to place it in an optional dependency group, I think `neo4j` would be a better fit.

You’re right that I had something else in mind for the `bigdata` group, but Parquet wasn’t actually one of them.

slobentanzer:
ah yeah, that makes sense. if it's tied to neo4j format, it can go in that group.

so `bigdata` is for the disk backed deduplication?

kpto:
@slobentanzer I was actually experimenting with a parallelisation idea, so bigdata was partly intended for that as well. But even for disk-backed deduplication alone, I think bigdata is a simple and easily understood group name that focuses on the intent rather than the implementation details.

Should Parquet become the default format?

slobentanzer:
Sounds good.

> Should Parquet become the default format?

I think so, yes. You mean for Neo4j, right? I always browse the CSVs to quickly check something, but that should also work with the dedicated CSV output format. Happy to move forward.

Is there something still missing from this PR now that we've made this decision?

kpto:
@slobentanzer Just a reminder, if Parquet becomes the default format, the optional `neo4j` dependency group will be required for offline Neo4j builds. Are you okay with that?

slobentanzer:
@kpto yes, I am fine with that. Neo4j should be an option like any other I think.

kpto:
@slobentanzer The `pyarrow` dependency has been moved to group `neo4j`, Parquet is now the default format, and the documentation has been updated.

Note that because ArangoDB inherits the Neo4j batch writer, I added a `writer_requested` argument to the `_BatchWriter` base class constructor (`__init__`). This allows the base class to determine which writer was actually requested by the upstream caller, so the Neo4j module defaults the output format to Parquet only when the requested writer is genuinely Neo4j.

kpto:
@slobentanzer I can see that changes have been requested on the PR, but I can’t see what changes are required.

slobentanzer:
I think that still refers to my review in https://github.com/biocypher/biocypher/pull/522#pullrequestreview-4367641993

Not sure how many of these are still relevant, anything up for discussion of course. Feel free to summarise from your own point of view and tag me for review once ready.

kpto:
@slobentanzer All issues should already have been resolved.

slobentanzer:
@kpto how does import call construction work after this PR?

1. only supported in Neo4j >= 5.26 via `--input-type=csv|parquet`; what happens when the user runs a lower version? We detect v4 vs v5, but not 5.25 vs 5.26.
1. how does this automatic detection work; does it require a change in the import call? I tried with Neo4j 5.26.28 (community) and it failed; `IMPORT FAILED ... Error in input data / Error importing csv file`. Adding the `--input-type` flag fixed that.

Also noted that the `removesuffix("[]")` header workaround appears unnecessary; where did you see that failure? Running  an example with `genes:string[]` in the header imported fine and gave the same graph as the stripped one. Maybe was patched?

We could add a test that checks the integration of the generated import script with Neo4j (like in the Postgres tests).

kpto:
@slobentanzer See https://neo4j.com/release-notes/database/neo4j-5-26-28/

Neo4j has improved Parquet support. The previous format detection behaviour was likely a bug rather than an intentional feature, so the import script needs to be updated accordingly. The calendar version that includes these fixes is believed to be 2026.06.0. I also suspect that the `:LABEL` column now supports arrays, though this still needs to be verified.

Given the current situation, we should align with the latest behaviour: revert the default format to CSV and document that Parquet should only be used with version 5.26.28/2026.06.0 or later.

slobentanzer:
@kpto but we do have version detection in the script already. Why not just make it work regardless? If the user has an earlier version of Neo4j, or chosen CSV on purpose, we use CSV. Otherwise, go to Parquet (that's the default you suggested, right?). Obviously, this applies only to new biocypher versions (likely >0.17) and Neo4j versions above the threshold that we set in the script.

kpto:
@slobentanzer It depends on whether we want to align with the new behaviour. If so, those versions would become the minimum requirement; since they were released only a month ago, BioCypher would no longer work automatically with any Neo4j version released before that. A user might upgrade BioCypher without also upgrading Neo4j.

slobentanzer:
@kpto I am not sure we are aligned on the issue. Is there a reason why the import script cannot do this dynamically?

Very simply put:
- default to Parquet (i.e., support the new behaviour)
- if any premise does not hold (e.g., older Neo4j, dedicated choice of CSV in the config), fall back to CSV

The import script already is conditional; you do know how that works, right?

> We could add a test that checks the integration of the generated import script with Neo4j (like in the Postgres tests).

This is probably already testable with the existing integration tests, we just need to extend them to the actually exported script.

kpto:
@slobentanzer I've been assuming you mean the generated script that runs the `neo4j-admin` command to import the tables. Or do you mean a different one that I'm not aware of?

Falling back in that script won't make the already-generated Parquet files importable into a Neo4j instance whose version doesn't support Parquet or the correct header format with `[]`.


To clarify, here’s what I mean:
1. The user runs BioCypher.  
2. BioCypher defaults to Parquet output.  
3. It produces Parquet tables, CSV headers, and an import script.  
4. The script detects a Neo4j version that does not support the correct header format, what can it do in that case?

slobentanzer:
Ah so you mean that we don't know at the beginning of the pipeline what the user will then run at import time. That's true.

So the choices would be to either print an error (causing the user to have to either update or re-run) or to make Parquet a deliberate choice for the time being. I suppose the latter, with a "deprecation note" of sorts, could be a good option? And then we switch to Parquet default in a couple months?

kpto:
@slobentanzer Maybe we just keep the current header and add the `--input-type` flag to the script. I’ve already added it.

slobentanzer:
@kpto how does this prevent the failure that you mentioned (user with older version not being able to import from Parquet)? Do we print a warning when biocypher starts with Parquet selected? What do you think about the deprecation-like message / migration strategy?

slobentanzer:
@kpto can older versions of Neo4j handle the input-type flag?

kpto:
@slobentanzer It doesn’t; my suggestion simply widens the supported version range. I also don’t know how we can emit an accurate warning, since we don’t know the exact minor and calendar version that first introduced Parquet support. We could print a generic warning that simply tells the user to ensure their Neo4j instance supports Parquet.

Regarding the deprecation-style message: rather than saying the default format *will* be changed, I’d prefer to state that the default format *has* been changed, and that users should either confirm their Neo4j instance supports Parquet or explicitly select CSV. That’s purely a subjective preference, of course.

I believe the `--input-type` flag has been present since Parquet support was added.

There’s also another option: default to Parquet only when running in `bigdata` mode.

slobentanzer:
> We could print a generic warning that simply tells the user to ensure their Neo4j instance supports Parquet.

agree

> I’d prefer to state that the default format has been changed, and that users should either confirm their Neo4j instance supports Parquet or explicitly select CSV

also agree

> I believe the --input-type flag has been present since Parquet support was added.

doesn't this mean that an old neo4j will throw an error when encountering the new import script? even if file type is set to CSV, if the CLI does not know the parameter, it will error out instead of ignoring, right? I am just assuming though.

if this is the case, we can't just add the param to the script flat out.

kpto:
@slobentanzer You're right, I overlooked that. I've updated the code so the flag is added only when the file format is Parquet, and I've also added a warning message reminding users to check their Neo4j version for Parquet support.

slobentanzer (dismissed):
@kpto thanks; happy now. I did some polishing and changed the test suite to community edition, which I wanted to do for a while.

Fine to merge, would appreciate a double check of the additions though.

kpto:
@slobentanzer A reminder, Neo4j now uses calendar versioning. The previous semantic versioning is reserved for LTS releases. Calendar versions are not the same as semantic versions, so Neo4j is retrofitting new features onto their LTS releases.

Additionally, the change marks 5.26.28 as the minimum required version, the one released only a month ago. Are you sure that’s what you want?

slobentanzer:
@kpto good point, thanks. I missed that we didn't revert to CSV default fully; now implemented.

I ran some tests on the recent versions, with some interesting results:
- the array workaround is necessary for 2026.03.1 and 2026.04.0
- CSV got a new problem in 2026.06.0; a new param `--vector-delimiter`, default `;`, that clashes with biocypher's default `--delimiter`. Using Parquet fixes it; need to check what to do about the CSV going forward. New issue created: https://github.com/biocypher/biocypher/issues/583

Was informative to run this matrix; the issue doesn't directly affect this PR, but will affect users that use the newest Neo4j version. You think we should fix before merging?

kpto:
@slobentanzer I don’t think Neo4j’s Parquet support is stable enough to be the default, it’s just that CSV is so terrible that I’m willing to give Parquet a chance.

Speaking only for myself, I find dealing with Neo4j version differences easier than wrestling with delimiters, especially since Neo4j overloads CSV to handle too many rich data types. I believe even the Neo4j team recognizes the problem and wants to move to a better format.

Potential compatibility issues with Parquet are easily fixable. Delimiter problems, on the other hand, affect every produced record; you can’t even easily identify which record is the source of the conflict, let alone work out how to resolve it.

On the delimiter issue itself, I think BioCypher should not try to solve Neo4j’s problems. It should simply defaults to whatever Neo4j defaults to.

slobentanzer:
@kpto agree, could you please have a quick look at #584 ? Might be useful to merge that first and then this one?

kpto:
@slobentanzer I don’t think the merge order matters much.

Also, I’m not entirely clear on the current decision, are you agreeing with everything in my last comment about making Parquet the default?

slobentanzer:
@kpto I think I didn't understand what you meant. "Parquet is not stable enough to be the default" for me was a clear indicator that you suggest reverting to CSV. You can still use parquet no problem even if it is not the default, so that wasn't a contradiction for me. Could you please restate what you think we should do in simple terms?

kpto:
@slobentanzer Sorry, I meant that Neo4j’s Parquet support isn’t stable enough yet, but I still want to make it the default. In my view CSV is unusable and not a viable solution, once a project moves beyond toy examples, delimiter problems appear quickly. As I explained in my previous comment, post-build remedies are possible with Parquet but not with CSV.

slobentanzer:
OK, feel free to revert this PR to the state you envision; I trust your assessment. Let me know when I should do the final check. I fully agree and have seen many delimiter problems from the start, so it's nice that we now have an alternative.

kpto:
@slobentanzer Actually, could you help me test something first? I’d like to know which versions need the [] workaround. You only mentioned 2026.03.1 and 2026.04.0, does that mean older versions are fine, or have they just not been tested yet?

My laptop isn’t in a state that makes running the matrix test easy right now.

slobentanzer:
which versions do you want to test? happy to throw it to claude. a list is fine.

kpto:
@slobentanzer 5.26.0, 5.26.2, 5.26.3, 2025.01.0, 2025.02.0, 2025.03.0, thanks.

slobentanzer:
@kpto ran the six you listed, and while the harness was set up I extended it across the whole range so the table is complete and measured the same way throughout. Short answer to your question: none of the six need the `[]` workaround, because none of them can import Parquet at all. The workaround is needed only on 2026.03.1 and 2026.04.0.

Setup: generated a real BioCypher Neo4j graph (Protein nodes with a `genes` string-array property plus one edge) with each community image, ran the generated `neo4j-admin` import call in the container, started the server and read the graph back with `cypher-shell`. Every row below is measured, not recalled. Three variants per version:

- **CSV** control.
- **Parquet, bare header** = current code, `genes` declared with no array type.
- **Parquet, `string[]` header** = the `removesuffix("[]")` workaround removed, `genes:string[]` declared.

| Neo4j | CSV import | Parquet, bare header (current) | Parquet, `genes:string[]` header | `[]` workaround needed? |
| --- | --- | --- | --- | --- |
| 5.26.0 | ok | fails to import | fails to import | n/a, no Parquet |
| 5.26.2 | ok | fails to import | fails to import | n/a, no Parquet |
| 5.26.3 | ok | fails to import | fails to import | n/a, no Parquet |
| 5.26.25 | ok | fails to import | fails to import | n/a, no Parquet |
| 5.26.26 | ok | ok | ok | no |
| 5.26.27 | ok | ok | ok | no |
| 5.26.28 | ok | ok | ok | no |
| 2025.01.0 | ok | fails to import | fails to import | n/a, no Parquet |
| 2025.02.0 | ok | fails to import | fails to import | n/a, no Parquet |
| 2025.03.0 | ok | fails to import | fails to import | n/a, no Parquet |
| 2026.03.1 | ok | ok | imports, **array → NULL** | **yes** |
| 2026.04.0 | ok | ok | imports, **array → NULL** | **yes** |
| 2026.05.0 | ok | ok | ok | no, fixed |
| 2026.06.0 | **fails, #583** | ok | ok | no, fixed |

`ok` for CSV and for a Parquet row means the import succeeded and both the `genes` array and the `score` scalar came back on readback. Legend for the failure cells:

- **fails to import**: `neo4j-admin` accepts `--input-type=parquet` (never rejected as unknown on any version), then hands the plain-text `.csv` header file to its Parquet reader, which rejects it: `... is not a Parquet file. Expected magic number at tail, but found [65, 66, 69, 76]`. Those bytes are ASCII `A B E L`, the tail of the header ending in `:LABEL`. Loud error, no data written.
- **array → NULL**: import reports IMPORT DONE, but `n.genes` reads back as `NULL` while `n.score` survives. This is the silent case the bare-header workaround exists for. Readback on 2026.03.1: `genes, score` / `NULL, 1.0`.
- **fails, #583**: `Character ';' specified by vector delimiter is the same as specified by delimiter`. CSV only, the vector-delimiter clash.

What this settles:

1. Parquet import works from **5.26.26** on the LTS line and from **2026.03.1** on the calendar line. Everything below either boundary fails loudly (the header-as-Parquet bug). The first three 2025 calendar releases behave exactly like the old 5.26.x line.
2. The `[]` workaround is needed only on **2026.03.1 and 2026.04.0**. On 5.26.26 through 5.26.28 both header forms import the array fine, and Neo4j fixed the native `string[]` handling by **2026.05.0**, so from there on the bare form is no longer required (still harmless).
3. CSV works on every version except **2026.06.0** (#583). Parquet is the only format that imports on 2026.06.0 with the stock delimiter.

Net for the default decision: on every version where Parquet is unsupported it fails loudly at write-import time, not silently. The only silent mode is the array → NULL on 2026.03.1 / 2026.04.0, which the bare-column workaround on this branch already prevents.

kpto:
@slobentanzer I decided to remove the header files and use column names directly from the Parquet files. This avoids a lot of problems and significantly widens the range of usable versions.

I also reverted the “make CSV default again” commit.

Would be great if you could run the matrix test again.

slobentanzer:
> the numbers below come from real experiments: actual `neo4j-admin` imports in Docker containers, each graph read back with `cypher-shell`. The write-up is AI-generated (Claude), posted by me.

@kpto I re-ran the version matrix against your parquet-based-header change (`b80a822e`, now the branch tip). The earlier table in this thread was for the separate-CSV-header approach, so it no longer describes the code. Here is how the current code behaves.

Setup: generated a real BioCypher Neo4j graph (Protein nodes with a `genes` string array, one edge), ran the generated `neo4j-admin` import call in each community image, started the server and read each node back **by its `id` property, not by `:Protein` label**, so a node whose `:LABEL` imported wrong is still seen rather than silently missed.

| Neo4j | CSV | Parquet (current code) |
| --- | --- | --- |
| 5.26.0 | ok | IMPORT DONE, but **labels corrupted (silent)** |
| 5.26.2 | ok | labels corrupted (silent) |
| 5.26.3 | ok | labels corrupted (silent) |
| 5.26.25 | ok | labels corrupted (silent) |
| 5.26.26 | ok | ok |
| 5.26.27 | ok | ok |
| 5.26.28 | ok | ok |
| 2025.01.0 | ok | labels corrupted (silent) |
| 2025.02.0 | ok | **hard crash, 0 nodes (NullPointerException)** |
| 2025.03.0 | ok | hard crash, 0 nodes (NullPointerException) |
| 2026.03.1 | ok | ok |
| 2026.04.0 | ok | ok |
| 2026.05.0 | ok | ok |
| 2026.06.0 | **fails, #583** | ok |

`ok` for a parquet row means IMPORT DONE, 2 nodes, the `Protein` label present, the `genes` array intact, `score` intact, and the relationship intact.

Three regimes for parquet on the current code:

1. **Fully working**: 5.26.26 / 5.26.27 / 5.26.28 on the LTS line, and 2026.03.1 / 2026.04.0 / 2026.05.0 / 2026.06.0 on the calendar line.
2. **Silent label corruption** (5.26.0, 5.26.2, 5.26.3, 5.26.25, 2025.01.0): import reports IMPORT DONE, node properties and relationships are correct, but the `:LABEL` value `Protein|Polypeptide|BiologicalEntity|NamedThing|Entity` is split into one label per **character** (`"a"`, `"B"`, `"P"`, `"|"`, ...) instead of per `|`. Readback of node `p1`: `labels = ["a","B","c","d","e","E","g","h","i","l","m","n","N","o","P","p","r","t","T","y","|"]`, `genes = ["g1"]`, `score = 1.0`. The node has no `Protein` label, so nothing finds it by label. The relationship `:TYPE` survives because it carries no `|`.
3. **Hard crash** (2025.02.0, 2025.03.0): `java.lang.NullPointerException at org.neo4j.internal.batchimport.input.parquet.ParquetInput.lambda$new$6`, 0 nodes imported.

## What the change bought, and the tradeoff

Background for anyone reading later: the previous parquet path wrote a separate `.csv` header file (column names and types) next to the `.parquet` data, and named the parquet columns positionally (`"0"`, `"1"`, ...). Neo4j versions below 5.26.26 fed that plain-text `.csv` header to their parquet reader and rejected it (`not a Parquet file. Expected magic number at tail`), so parquet failed loudly on every unsupported version. This change drops the header, names the parquet columns with the real Neo4j names (`:ID`, `genes`, `:LABEL`, `:START_ID`, `:TYPE`, ...), and lets types come from the parquet schema.

Two clear wins:

- **The `[]` array workaround is gone and no longer needed.** With the old header, declaring an array as `x:string[]` over a native parquet list imported the array as NULL on 2026.03.1 and 2026.04.0, which is why the bare-column workaround existed. With no header, the array is a native list column and survives on those versions natively.
- The working set is unchanged: parquet still imports correctly on exactly 5.26.26+ and 2026.03+.

The cost:

- **On unsupported old versions the failure went from loud to silent for part of the range.** Old approach: every unsupported version failed the import outright, no data written. Current approach: 5.26.0 / 5.26.2 / 5.26.3 / 5.26.25 / 2025.01.0 report IMPORT DONE and produce a graph that looks populated but whose node labels are destroyed. 2025.02.0 and 2025.03.0 still crash loudly. Since parquet is now the default, a user on Neo4j older than 5.26.26 (LTS) or on an early 2025 calendar release who does not act on the startup warning gets a silently broken graph rather than an error.

The `Protein|Polypeptide|...` label string is BioCypher's own encoding of the `:LABEL` column (`|`-joined ancestry). Neo4j 5.26.26+ and 2026.03+ split it on the array delimiter correctly; the pre-fix versions iterate it per character. One option, if we want to reduce the blast radius on old versions: write `:LABEL` in parquet as a native list column rather than a `|`-joined string, and check whether that avoids the per-character split or at least turns it into a loud failure. I have not tested that yet.

Net: the change is the right direction and strictly better on supported Neo4j. The one thing to weigh before shipping parquet as the default is the silent label corruption on 5.26.0 through 5.26.25 and 2025.01.0, since a silent wrong graph is worse to debug than a hard error. The startup warning naming the 5.26.26 / 2026.03 floor is the current mitigation.

kpto:
@slobentanzer I wasn’t able to reproduce the error you encountered, I tested on both 2025.01.0 and 2025.02.0 and everything worked fine.

That said, I suspect the non-default delimiter setting was the cause. I’ve updated the code so that delimiter and quote settings are ignored and removed from the import script whenever Parquet is used. Could you test it again?

Also, as far as I know, the :LABEL column does not support the native array type.

slobentanzer:
> Real experiments (actual `neo4j-admin` imports in Docker, each graph read back with `cypher-shell`). Write-up is AI-generated (Claude), posted by me.

@kpto Re-ran against the current tip (`1edd2d12`). Your fix works: the label corruption is gone, and the usable range is much wider now.

| Neo4j | CSV | Parquet |
| --- | --- | --- |
| 5.26.0 | ok | ok |
| 5.26.2 | ok | ok |
| 5.26.3 | ok | ok |
| 5.26.25 | ok | ok |
| 5.26.26 | ok | ok |
| 5.26.27 | ok | ok |
| 5.26.28 | ok | ok |
| 2025.01.0 | ok | ok |
| 2025.02.0 | ok | **fails** |
| 2025.03.0 | ok | **fails** |
| 2026.03.1 | ok | ok |
| 2026.04.0 | ok | ok |
| 2026.05.0 | ok | ok |
| 2026.06.0 | fails (#583) | ok |

`ok` = IMPORT DONE, 2 nodes, correct labels, `genes` array intact, `score` intact, relationship intact.

Two points:

1. You were right about the delimiter. `:LABEL` joined with `;` and no `--array-delimiter` in the call imports labels correctly on the versions that previously split them per character (5.26.0, 5.26.2, 5.26.3, 5.26.25, 2025.01.0). No silent corruption remains anywhere: every failure now is a loud abort with 0 nodes.

2. The 2025.02.0 / 2025.03.0 crash is real and is **not** the delimiter. The trigger is a native array (list) column. Minimal reproducer that crashes 2025.02.0 and 2025.03.0 but imports fine on 2025.01.0, all 5.26.x, and 2026.03+:

```python
import pyarrow as pa, pyarrow.parquet as pq
pq.write_table(pa.table({":ID": ["p1"], "genes": [["g1"]], ":LABEL": ["Protein"]}), "n-part000.parquet")
# neo4j-admin database import full neo4j --input-type=parquet --overwrite-destination=true --nodes="n-part.*"
# 2025.02.0 / 2025.03.0: NullPointerException at ParquetInput.lambda$new$6, 0 nodes
```

Drop the `genes` list column and it imports fine. So Neo4j's parquet importer regressed for native list columns between calendar 2025.01 and 2025.02, and was fixed again by 2026.03. That is likely why you couldn't reproduce it: the test graph needs a list-typed property.

Also worth updating: the startup warning still says the floor is "5.26.26+ on LTS, or 2026.03+". With this change that is too conservative and slightly wrong. Measured: the whole 5.26 line works (5.26.0 up), 2025.01.0 works, and the broken spot on the calendar line is 2025.02.0 / 2025.03.0 specifically, not a floor. I have no images for 2025.04 through 2026.02, so that stretch is untested.

kpto:
@slobentanzer We can’t guarantee the correctness of those messages given the combination of data content, platform, and versions, so I’d suggest not mentioning versions in the messages, otherwise we risk taking on unnecessary responsibility. That said, I’ll leave the final call to you.

Apart from the messaging issue, are you happy with the PR?

slobentanzer:
@kpto thanks; can you clarify: what are "these messages"?

kpto:
@slobentanzer

> Also worth updating: the startup warning still says the floor is "5.26.26+ on LTS, or 2026.03+". With this change that is too conservative and slightly wrong. Measured: the whole 5.26 line works (5.26.0 up), 2025.01.0 works, and the broken spot on the calendar line is 2025.02.0 / 2025.03.0 specifically, not a floor. I have no images for 2025.04 through 2026.02, so that stretch is untested.

slobentanzer:
@kpto fine to go with your opinion here. I think we should make the users aware that there can be a problem, and it needs to be at runtime, not just somewhere in the docs. Beyond that, fine to go with your recommendation.

kpto:
@slobentanzer I’ve removed the mentions of specific versions. Could you take a look?

slobentanzer (approved):
@kpto thanks and looks good now on my end!

PR #528, opened 2026-05-25
Add ontoweaver tutorial

yasfazl:


slobentanzer:
@yasfazl thanks! Please merge current main at your convenience; there were conflicting changes to the version you based on. Please make sure to keep current main functionality intact.

slobentanzer:
Also, please add a high-level description of your PR so I know what your aims are with it.

jdreo:
If it's not too much of a burden, I'll suggest splitting the tutorial in subfiles, so that it is possible to automatically extract the parts which source code can be turned into a test. 

Then add a test that does that, so that you can automatically double-check that the tutorial is up-to-date.

There's an example in the OntoWeaver project doing that (albeit using rST and not Markdown): https://github.com/oncodash/ontoweaver/blob/main/tests/test_full_fledged_demo.py

slobentanzer:
@jdreo good point; we have doctest as a dependency if I remember correctly. That should allow you to do this @yasfazl

PR #537, opened 2026-05-28
feat(deduplicate): disk backed deduplicator for large scale datasets

kpto:
### Problem

The current deduplication implementation stores all seen entity and relationship IDs in Python sets. This approach is memory-intensive and does not scale well because memory usage grows unbounded with dataset size. Large datasets can easily exceed the memory capacity of a typical laptop.

### Improvement

A disk-backed deduplicator has been implemented. It is fully compatible with the existing implementation and can be swapped in transparently to enable large-scale processing.

The new deduplicator uses a three-layer lookup mechanism:

1. An in-memory pending Python set
2. An in-memory Bloom filter acting as a fast probabilistic gate
3. An on-disk LMDB database as the authoritative store

The pending set is flushed into LMDB once it exceeds 100,000 items, keeping memory usage bounded while avoiding frequent LMDB writes, which are relatively expensive.

The Bloom filter uses a fixed amount of memory. It guarantees absence checks while allowing a small probability of false positives. With the default configuration, it allocates approximately 2.8 GB of memory and achieves a false-positive rate of 0.00001 for 1 billion items. Assuming all IDs are unique, this results in roughly 10,000 on-disk LMDB validation lookups caused by Bloom filter false positives.

All IDs are normalised using `xxhash128`. For 1 billion IDs, the collision probability is negligibly small in practice. Each hash value uses roughly the same memory as a six-character Python string. `xxhash` was chosen because it is specifically optimised for high-speed hashing workloads such as this one.

Entity and relationship IDs are namespaced before hashing and stored in the same index.

### Specification

1. Added a new base configuration parameter: `big_data`.
2. `big_data` is only supported in `offline` mode; otherwise a `ValueError` is raised.
3. When `big_data=True`, the core instantiates the disk-backed deduplicator instead of the original implementation.
4. Disk-backed deduplicator behaviour:

   * Membership checks follow the sequence: pending set → Bloom filter → LMDB.
   * Bloom filter negatives short-circuit the lookup.
   * Bloom filter positives are validated against LMDB.
   * Namespace separation prevents entity and relationship key collisions.
   * The pending set is automatically flushed once the batch threshold is reached.
5. Added optional dependencies under the `bigdata` extra:

   * `xxhash`
   * `lmdb`
   * `rbloom`

### Other Changes

The original deduplicator implementation was slightly cleaned up, mainly through variable renaming, while preserving its behaviour.

One detail remains unclear to me: relationship IDs are currently disjointed by relationship type. In Neo4j offline import mode, this may allow edges with the same ID to coexist, which Neo4j may interpret as invalid input data.

### Remarks

I am the developer of the Open Targets adapter, and Open Targets datasets operate at billion-scale. This change is motivated directly by the need to process datasets of that size efficiently.

This proposal is part of a broader effort to make BioCypher capable of handling large-scale data processing workloads. For that reason, the configuration parameter is named `big_data`, reflecting the possibility that additional large-scale optimisations may be introduced under the same feature flag in the future.

Documentation updates will be added after feedback on the feature has been received.


slobentanzer (commented):
@kpto thanks, really nice feature. Agree that it is a bottleneck, and also agree with the implementation strategy. Just some minor comments:

## edge types

> In Neo4j offline import mode, this may allow edges with the same ID to coexist, which Neo4j may interpret as invalid input data.

Did you check? Neo4j always creates its own index, and relationships in particular are not under many constraints in the general setup, so it would surprise me if it were a problem. Should be easy to check.

The main reason that we do deduplication is biocypher-internal; that we don't want to end up with several edges of the same type between the same two nodes. But for that, a smart choice of edge ID generation is equally important (it's just not related to Neo4j processing). The reason we leverage type here is that in biology and many other fields, relationships don't have mature ID systems like entities. The main risk is deduplicating edges that are not actually duplicates if type information is not present. E.g., protein A `interacts_with` protein B and protein A `regulates` protein B both would synthesise to edge id `A_B` (if you use naive concat, or also hash from the same properties).

This is an issue beyond Neo4j, and also will apply to the non-big-data deduplicator; I think the only reasonable approach is to use global ID for nodes and typed ID for edges, consistently everywhere. Happy for counterarguments though.

This applies to reified edges as well, a fix would be to prefix with type in `rel_as_node_seen` too (`f"{relationship_type}_{relationship_id}"`).

I think there are no tests parameterising `rel_as_node_seen` or the difference between regular and reified nodes. Could be considered for addition.

## duplicate logging

`Deduplicator.get_duplicate_nodes()` / `get_duplicate_edges()` previously returned `None` when there were no duplicates; the refactor now always returns `(set, set)`. In biocypher/_core.py:638-678:

```
dn = self._deduplicator.get_duplicate_nodes()
if dn:
      ...
else:
      logger.info("No duplicate nodes in input.")
```

`bool((set(), set()))` is True (non-empty tuple), so the else branch never executes. Every run will now log a header like "Duplicate node types encountered (IDs in log):\n" with no body, and the same for edges. Solutions would be to either restore the None return when sets are empty, or change the call sites to check if dn[0] or dn[1].

I don't think any test captures that yet.

## notes

- DiskBasedDeduplicator still inherits the in-memory `duplicate_entity_ids` / `duplicate_relationship_ids` sets from `Deduplicator.__init__()`. If duplicates are common, those sets can also grow unbounded.
- We would allocate almost 3GB the instant a user passes `big_data=True`. We could instead use a lower default (e.g., 100MB / 1000MB) and expose the knobs via base_config rather than only via constructor args.
- If `lmdb_path` is user-supplied and persisted, a crash mid-build leaves a corrupt/partial index; could be worth a note in the docs.
- Should we add `pyarrow` to the big_data group?

kpto:
@slobentanzer Regarding edge ID conflicts, you're right, I had forgotten that neo4j-admin import does not use the edge IDs you provide.

I would still suggest handling relationship namespacing through ID generation rather than through deduplication logic. That way, users remain free to define their own notion of duplication when they supply relationship IDs themselves.

```python
relationship_id = relationship.get_id() or (
    f"bc_{relationship_type}_{relationship.get_source_id()}_{relationship.get_target_id()}"
)
```

This should already solve the issue. The `bc` prefix helps minimise conflicts with user-supplied IDs while signalling that the ID was generated by BioCypher.

I've fixed the duplicate logging, although I'm not convinced that unit-testing log messages is a good idea. It feels a bit unusual to me.

The current implementation assumes reasonably clean input data with limited duplication. If that assumption does not hold, the first symptom you'll likely notice is severely degraded performance, as the system will repeatedly hit the disk-backed LMDB before memory becomes exhausted by the in-memory sets.

At the moment, there is no way to specify the LMDB path, so the database is always created and discarded automatically. I'm also unsure whether exposing those controls, including Bloom filter parameters in the BioCypher configuration is worthwhile, as it would add complexity to both the interface and the documentation.

The PR has also been updated to incorporate the latest changes from the main branch.

slobentanzer (dismissed):
@kpto similar to my question in the other PR, anything left to do for this one before the merge? is the `bigdata` concept set up to receive more features seamlessly in the future?

kpto:
@slobentanzer I can’t be sure whether `bigdata` will receive more features in the future, but I think it is a very good fit for this feature nonetheless.

I’ve added documentation. Most of it was AI-generated and then reviewed. Please take a look.

kpto:
@slobentanzer Any update?

slobentanzer:
@kpto thanks for the nudge; was busy / summertime

slobentanzer:
@kpto functionally, there is still this regression (previous code will not deduplicate, but this version does). From my comment

> This applies to reified edges as well, a fix would be to prefix with type in rel_as_node_seen too (f"{relationship_type}_{relationship_id}").

and your response

> I would still suggest handling relationship namespacing through ID generation rather than through deduplication logic. That way, users remain free to define their own notion of duplication when they supply relationship IDs themselves.
    ```
    relationship_id = relationship.get_id() or (
        f"bc_{relationship_type}_{relationship.get_source_id()}_{relationship.get_target_id()}"
    )
    ```

→ none of the two is implemented, so a user could get a different graph after this merge and using the `bigdata` mode

kpto:
@slobentanzer Thanks, I overlooked relationship as node. That should now be fixed, and I’ve also added tests for it.

slobentanzer (approved):
@kpto great, many thanks. ready to merge; let me know if I should do it or go ahead.

PR #539, opened 2026-05-29
feat(batch_writer): stream edge processing in offline mode to eliminate unbounded memory usage

kpto:
### Problem

The current edge-writing implementation forces exhaustive materialisation of generators for two reasons:

1. To determine whether the generator is empty.
2. To collect all nodes and edges produced when resolving relationships-as-nodes.

This design results in unbounded memory usage, limiting BioCypher's ability to scale to large datasets.

### Improvement

For the emptiness check, a simple boolean flag set during iteration is sufficient. Although evaluated repeatedly, the overhead is negligible in practice and compares favourably to alternative approaches in both performance and code simplicity.

For relationships-as-nodes processing, entities are now resolved incrementally rather than accumulated upfront. Each resolved node and edge is passed directly to downstream processing as soon as it is produced, eliminating the need to store the entire result set in memory.

To support this, `_write_node_data` and `_write_edge_data` have been refactored and renamed. Conceptually, they now operate in a streaming fashion. Instead of processing an entire collection of nodes or edges, each function returns two handles (closures):

* One for processing individual items as they are produced.
* One for flushing any remaining buffered data at the end of the sequence.

This allows callers to stream resolved nodes and edges immediately without intermediate accumulation. Wrapper functions preserve the original `_write_node_data` and `_write_edge_data` interfaces, maintaining full backwards compatibility.

### Other Changes

The original implementation accepted `Generator` as the input type. Since the input is only iterated over and does not rely on generator-specific behaviour, `Iterable` is a more appropriate and less restrictive interface. The relevant type annotations have therefore been updated accordingly.

### Remarks
Testing on the Open Targets dataset shows that memory usage stabilises with this optimisation in place. Combined with [https://github.com/biocypher/biocypher/pull/537](https://github.com/biocypher/biocypher/pull/537), BioCypher should now be able to process datasets of arbitrary size without memory consumption growing with the total input size.

As this change only affects implementation details and does not modify external behaviour, no additional tests should be required.

### DIFF
--- a/biocypher/output/write/_batch_writer.py
+++ b/biocypher/output/write/_batch_writer.py
@@ -6,11 +6,9 @@ import re
 
 from abc import ABC, abstractmethod
 from collections import OrderedDict, defaultdict
-from types import GeneratorType
+from collections.abc import Iterable
 
 import networkx
-
-from more_itertools import peekable
 
 from biocypher._create import BioCypherEdge, BioCypherNode, BioCypherRelAsNode
 from biocypher._deduplicate import Deduplicator
@@ -355,7 +353,7 @@ class _BatchWriter(_Writer, ABC):
             logger.error(msg)
             raise ValueError(msg)
 
-        elif self.labels_order != "None":
+        if self.labels_order != "None":
             if self.edge_labels_order != self.labels_order or self.node_labels_order != self.labels_order:
                 msg = (
                     f"`labels_order`=`{self.labels_order}` "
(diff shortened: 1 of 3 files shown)


slobentanzer (dismissed):
Hi @kpto, great analysis and very nice solution. This has been bugging me for a long time.

I have only adjusted some names for more intuitive understanding and fixed some typos.

I also fixed a small bug that happened when all edges were deduplicated.

About the merging of this and the other two (parquet, deduplication): do you want to bring them together once all are ready, or do you want to merge separately as soon as they're ready?

No real difference there, as we can wait with release-please merging to bump the version number as soon as we're ready at any point. We could either do them one-by-one, or merge them as one "big data" feature together.

kpto:
@slobentanzer I prefer merging this PR as soon as it's ready. It's easier for me to update the other branches. Everything looks good to me.

One small complaint is that the docstrings in the tests are tied to a specific development event, which I see as an unnecessary knowledge leak. It makes the tests harder for future maintainers to understand and update, since it's difficult to verify the statement or assess whether that event is still relevant.

AI tends to inject session information into code, which I hate, but it's a minor issue so it's up to you. Feel free to press the button.

slobentanzer:
Agree and wasn't planning to keep it in there; was just for your information. I'll remove it and then merge.

slobentanzer (approved):

PR #555, opened 2026-06-11
feat(write): split nodes with varying property sets into separate header groups

slobentanzer:
## Summary

The CSV / Neo4j admin-import batch writer derives **one header per node label**, fixed from the first node it sees (or from the schema). When a later node of the same label carries a different property set, `_write_single_node_list_to_file` raises a *"has more or fewer properties than another"* error.

That is fine for schema-defined data, but it breaks **schemaless / optional-property input** — e.g. FHIR resources where only some patients carry a given attribute. Today the only ways around it are to pre-scan the entire input to learn the full property union, or to rewrite previously written part files whenever a new property appears. Both are expensive (a second pass over the source, or unbounded I/O re-writes).

This PR groups nodes by their **property signature** instead:

- Nodes that share the same property set *and types* go to the same group (header + part files + import-call entry); a new signature opens a new group.
- Neo4j admin import already accepts **multiple file groups for the same label**, so each group is emitted as its own `--nodes=` entry and properties stay naturally sparse across the graph.
- It's a single streaming pass — no pre-scan of the input, no rewriting of prior parts.

## Behaviour

- **Fully backward compatible, no user action.** Schema-configured or otherwise homogeneous data produces exactly one signature, so the group keeps the plain label and you get the *same* single header, part files and import call as before. All 44 pre-existing `test_neo4j.py` tests pass unchanged.
- Only previously-**erroring** (divergent) input changes: it now succeeds, producing additional suffixed groups, e.g. `Patient-header.csv` + `PatientGroup1-header.csv`.
- The first signature per label keeps the legacy file names; subsequent signatures get a file-name-safe suffix.
- A reappearing signature is routed back to its existing group (it does not spawn a new one, and appends a new part file across calls).
- The real ontology `:LABEL` is preserved for every node regardless of group — the internal group key never leaks into the data.

## How it works

In `_BatchWriter` (writer-agnostic), the node-write path now:
1. determines each node's property/type dict (`_get_node_property_types`, extracted from the existing inline logic), and
2. maps its order-invariant, type-aware signature to a group key (`_get_node_group_key`).

All existing per-label bookkeeping (`bins`, `parts`, `node_property_dict`, `import_call_nodes`) is keyed by this group key. The Neo4j header/import-call code needed **no change** — it already iterates per entry and supports multiple `--nodes=` groups for one label.

## Scope & known follow-ups (intentionally out of this PR)

- **Nodes only.** Edges retain their current behaviour (their header writer does schema lookups by the dict key, so grouping them is a separate, larger change).
- **Memory** stays bounded per group by `batch_size`, as before — now per *signature* rather than per *label*. For input with a very large number of distinct property sets, a global cap on total buffered lines would be a sensible follow-up.
- A property that is consistently `None` forms its own signature (header column present, values empty). Harmless, but could be normalised later to reduce over-splitting.

## Tests

`test/output/write/graph/test_neo4j.py`, 11 new tests (full suite green, 69 passed / 4 DB-skipped):
- varying property sets split into groups, with matching headers/parts and both in the import call;
- homogeneous input still produces one legacy-named group;
- schema-1 / schema-2 / schema-1 reappearance, both within one call and across calls (part-index continuity);
- order-invariant grouping; type-difference splitting; multiple incrementing groups; independent grouping across labels;
- `:LABEL` preserved / group key never leaks; per-group batch flushing; `None`-valued property edge case.

## Review request

The MeDaX team — could you review this and, importantly, **try it on your actual FHIR data before we merge**? The motivating case is the patient-property divergence you hit; I'd like to confirm the generated headers, part files and `neo4j-admin` import call behave as expected on a real, large run (memory profile included). Not for merge until that's validated.

### DIFF
--- a/biocypher/output/write/_batch_writer.py
+++ b/biocypher/output/write/_batch_writer.py
@@ -310,6 +310,14 @@ class _BatchWriter(_Writer, ABC):
 
         self.parts = {}  # dict to store the paths of part files for each label
 
+        # map of node label -> {property signature: group key}. Nodes of a
+        # label that present different property sets (e.g. schemaless input
+        # with optional properties) are split into separate output groups so
+        # that each part file conforms to its own header. The first signature
+        # encountered keeps the plain label as its group key (i.e. legacy file
+        # names and import call); further signatures get a suffixed key.
+        self._node_property_groups = defaultdict(dict)
+
         self._labels_orders = ["Alphabetical", "Ascending", "Descending", "Leaves", "None"]
         self.labels_order = labels_order
         self.node_labels_order = node_labels_order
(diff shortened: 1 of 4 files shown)


slobentanzer:
@kpto FYI; I don't assume it conflicts with your proposed changes in the two open PRs, but I might miss something.

Gebbi8:
@slobentanzer : Greetings from MeDaX. I tested the feature branch against mock data that triggers our use case and cannot confirm the expected behaviour. The test runs two batches with Encounters that have different property signatures. Also the initial schema lacks some properties for Encounters initially. In the second batch the schema gets updated automatically towards the signature of the of the new Encounter.
At the end BioCypher  produces two csv tables for Encounters with different property signatures but neither a second header with '-Group' is created nor does the name of the second table has "-Group" in its name.

Gebbi8:
> @slobentanzer : Greetings from MeDaX. I tested the feature branch against mock data that triggers our use case and cannot confirm the expected behaviour. The test runs two batches with Encounters that have different property signatures. Also the initial schema lacks some properties for Encounters initially. In the second batch the schema gets updated automatically towards the signature of the of the new Encounter. At the end BioCypher produces two csv tables for Encounters with different property signatures but neither a second header with '-Group' is created nor does the name of the second table has "-Group" in its name.

After having a second look, I think that the feature should generally behave like it's schemaless part. How about this approach?

1. parse all property signatures for existing headers of the node type (maybe even all node types): allSig
2. FOR each node
    1. get property signature: nodeSig
    2. IF nodeSig already exists in allSig THEN  create new table in same group
    3. ELSE
        1. create new group header
        2. create new table in new group
        3. add new signature to allSig

slobentanzer:
@Gebbi8 does the new implementation not already do that? Or can you provide a minimal example where it breaks?

Naively, I would say this is exactly how the logic in this PR should work.

Are you specifying the properties in the schema as well as expecting them to be dynamic (at the same time)?

Gebbi8:
@slobentanzer I'm pretty sure it doesn't do that. And yes, we start with a existing schema for batch1, if this batch has new nodes, properties or edges, we extend the schema accordingly and then give the nodes, edges and the extended schema to Biocypher. This is repeated for every batch. So the schema is potentially updated in every batch before nodes and edges are generated.

For our use case the signatures of all nodes and groups would have to be parsed from the existing headers on the disk.

Gebbi8:
In which form would a minimal example be helpful for you? Would a test be sufficient or do you need a minimal version of our pipeline, so docker compose that creates a blaze server, python container and neo4j container?

slobentanzer:
@Gebbi8 so you're saying the schema can be modified during processing in any way? You can add a new node type, a new edge type, new properties, at any point?

> Would a test be sufficient

Yes, I am struggling with seeing the full extent of your problem. The minimal condition (full reflection of your workflow, not just a part) that leads to failure of the current biocypher codebase. If I can reproduce your failure, I can understand what changes would be needed. DevOps is not relevant, only the logic.

Gebbi8:
[test_schema_change_between_batches.py](https://github.com/user-attachments/files/29851413/test_schema_change_between_batches.py)

@slobentanzer This test reproduces our possible use case in a minimal way.



> @Gebbi8 so you're saying the schema can be modified during processing in any way? You can add a new node type, a new edge type, new properties, at any point?

Yes, in every batch - before the data is handed BioCypher - the schema might be updated if the new data has additional node types, edge types or properties.

slobentanzer:
@Gebbi8 could you please check if the fix I just applied makes it usable in your setting?

Gebbi8:
> @Gebbi8 could you please check if the fix I just applied makes it usable in your setting?

@slobentanzer thank you for the update. It is indeed usable now for our purpose. And I would be fine to us it that way.

Gebbi8:
Hey @slobentanzer, is there a schedule on when this feature will be added to the main branch? I would like to update our pipeline release soon to have it ready for the publication.

PR #562, opened 2026-06-16
fix(workflow): clear _seen_nodes/_seen_edges on remove/clear/from_json

### LINKED ISSUE #561: [BUG]: BioCypherWorkflow deduplication tracking not updated on remove_node, remove_edge, or clear
### Describe the bug

`BioCypherWorkflow` maintains two internal sets — `_seen_nodes` and `_seen_edges` — that track which entities have been added, so that the `deduplication=True` mode can prevent re-insertion of the same ID. However, three mutation methods fail to keep these sets in sync with the underlying graph:

1. **`remove_node(node_id)`** — removes the node from the graph but leaves `node_id` in `_seen_nodes`. The node can never be re-added afterwards.
2. **`remove_edge(edge_id)`** — same problem for `_seen_edges`.
3. **`clear()`** — replaces the graph with a new empty `Graph`, but **does not** clear `_seen_nodes` or `_seen_edges`. After a full reset, no previously-seen node or edge ID can ever be re-inserted.

An additional related gap: **`from_json()`** replaces the graph without populating `_seen_nodes`/`_seen_edges`, so loaded entities are invisible to deduplication checks.

### Steps to reproduce

```python
from biocypher import create_workflow

# --- remove_node ---
wf = create_workflow("test", deduplication=True)
wf.add_node("p1", "protein")
wf.remove_node("p1")
result = wf.add_node("p1", "protein")   # Expected: True — node was removed
print(result)                            # Actual:   False

# --- clear ---
wf2 = create_workflow("test2", deduplication=True)
wf2.add_node("p1", "protein")
wf2.clear()
result2 = wf2.add_node("p1", "protein")  # Expected: True — graph was cleared
print(result2)                            # Actual:   False
```

### Impact

Any workflow that removes or resets nodes/edges and then re-inserts entities with the same IDs will silently drop those insertions when `deduplication=True`, without any error or warning (unless `validation_mode="strict"` is also set).

### Python version

Any

### BioCypher version

current `main`

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

slobentanzer:
## Summary

Closes #561

`BioCypherWorkflow` tracks seen entity IDs in `_seen_nodes` and `_seen_edges` to power the `deduplication=True` mode, but three mutation methods failed to keep those sets in sync with the underlying graph:

- `remove_node` / `remove_edge` left stale entries, so a removed entity could never be re-added.
- `clear()` replaced the graph but left both sets populated, making it impossible to re-add any previously-seen entity after a full reset.
- `from_json()` replaced the graph without repopulating the sets, so newly-loaded entities were invisible to future deduplication checks.

### Changes

| Method | Fix |
|--------|-----|
| `remove_node(id)` | Discards `id` from `_seen_nodes` on success |
| `remove_edge(id)` | Looks up the edge type before removal; discards `(id, type)` from `_seen_edges` on success |
| `clear()` | Calls `_seen_nodes.clear()` and `_seen_edges.clear()` after replacing the graph |
| `from_json(json)` | Repopulates `_seen_nodes` and `_seen_edges` from the loaded graph |

## Test plan

- [x] `test_clear_resets_deduplication_tracking` — adds nodes/edges, clears, then re-adds the same IDs and asserts they succeed.
- [x] `test_remove_node_clears_deduplication_tracking` — adds a node, removes it, re-adds and asserts success.
- [x] `test_remove_edge_clears_deduplication_tracking` — adds an edge, removes it, re-adds and asserts success.
- [x] `test_from_json_populates_deduplication_tracking` — serialises a workflow, loads into a new deduplication-enabled workflow, and asserts that loaded IDs are treated as duplicates.
- [x] All 84 existing `test_workflow.py` tests continue to pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_016UqGDQ5Zyss4FZ5Ypau1Xq)_

### DIFF
--- a/biocypher/_workflow.py
+++ b/biocypher/_workflow.py
@@ -205,7 +205,10 @@ class BioCypherWorkflow:
         Returns:
             bool: True if node was removed, False if not found
         """
-        return self.graph.remove_node(node_id)
+        result = self.graph.remove_node(node_id)
+        if result:
+            self._seen_nodes.discard(node_id)
+        return result
 
     # ==================== EDGE OPERATIONS ====================
 
@@ -311,7 +314,11 @@ class BioCypherWorkflow:
         Returns:
             bool: True if edge was removed, False if not found
         """
-        return self.graph.remove_edge(edge_id)
+        edge = self.graph.get_edge(edge_id)
+        result = self.graph.remove_edge(edge_id)
+        if result and edge is not None:
+            self._seen_edges.discard((edge_id, edge.type))
+        return result
 
     # ==================== HYPEREDGE OPERATIONS ====================
 
@@ -619,6 +626,8 @@ class BioCypherWorkflow:
         data = json.loads(json_data)
         self.graph = Graph.from_dict(data)
         self.name = self.graph.name
+        self._seen_nodes = set(self.graph._nodes.keys())
+        self._seen_edges = {(edge.id, edge.type) for edge in self.graph._edges.values()}
 
     def save(self, filepath: str) -> None:
         """Save the graph to a file.
@@ -646,6 +655,8 @@ class BioCypherWorkflow:
     def clear(self) -> None:
         """Clear all nodes and edges from the graph."""
         self.graph = Graph(name=self.name, directed=self.graph.directed)
+        self._seen_nodes.clear()
+        self._seen_edges.clear()
         logger.info("Graph cleared")
 
     def copy(self) -> "BioCypherWorkflow":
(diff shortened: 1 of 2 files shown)


slobentanzer:
Closing in favor of #571, which fixes the same four methods (`_remove_node`, `remove_edge`, `from_json`, `clear`) but additionally guards the tracking-set updates behind `if self.deduplication:` — i.e. #562's changes are a subset of #571's, with #571 avoiding unnecessary work when deduplication is disabled. Superseded by #571.

PR #572, opened 2026-06-22
fix(neo4j): CSV escaping quote character

VladSam42:
- fixed special character escaping
- relaxed optional neo4j dependency to prevent conflict with other libs

### DIFF
--- a/biocypher/output/write/_batch_writer.py
+++ b/biocypher/output/write/_batch_writer.py
@@ -819,7 +819,7 @@ class _BatchWriter(_Writer, ABC):
                     elif isinstance(p, list):
                         plist.append(self._write_array_string(p))
                     else:
-                        plist.append(f"{self.quote}{p!s}{self.quote}")
+                        plist.append(self._quote_string(str(p)))
 
                 line.append(self.delim.join(plist))
             line.append(labels)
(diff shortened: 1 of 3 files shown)


slobentanzer:
Hi @VladSam42, which other libs require neo4j v4? They should probably update. Not sure we should simply relax in this case.

VladSam42:
biochatter depends on the abandoned neo4j-utils that enforces neo4j>=4.4,<5.0:
```
❯ uv sync --extra dev
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.12.*' and sys_platform == 'win32'):
  ╰─▶ Because biochatter depends on neo4j-utils>=0.0.12 and neo4j-utils==0.0.12 depends on neo4j>=4.4,<5.0, we can conclude that biochatter[podcast] depends on neo4j>=4.4,<5.0.
      And because biocypher[neo4j] depends on neo4j>=5.0, we can conclude that biocypher[neo4j] and biochatter[podcast] are incompatible.
      And because your workspace requires biochatter[podcast] and biocypher[neo4j], we can conclude that your workspace's requirements are unsatisfiable.
```

slobentanzer:
Let's remove or update in `biochatter` then. :)

slobentanzer:
@VladSam42 I have migrated the abandoned code into `biocypher` a while ago; it is likely a legacy issue / oversight on my part that we still depend on it there. Should be an easy fix (just use the driver code that is now in `biocypher`).

slobentanzer (approved):
Looks good, thanks!

Issue #575, opened 2026-06-25
[BUG]: `sqlite` section in default config is missing `labels_order` keys, making `BioCypher(dbms="sqlite")` always crash


slobentanzer:
### Describe the bug

`biocypher/_config/biocypher_config.yaml` has a `sqlite` section that is missing the `labels_order`, `node_labels_order`, and `edge_labels_order` keys that `_BatchWriter._check_labels_order()` requires to be non-`None`.

When `get_writer("sqlite", ...)` is called, it reads the config with:

```python
dbms_config = _config("sqlite") or {}
# ...
labels_order=dbms_config.get("labels_order"),       # → None  ← MISSING
node_labels_order=dbms_config.get("node_labels_order"),  # → None  ← MISSING
edge_labels_order=dbms_config.get("edge_labels_order"),  # → None  ← MISSING
```

These `None` values are passed to `_BatchWriter.__init__`, overriding the parameter defaults.  `_check_labels_order()` then immediately raises:

```
ValueError: A batch writer `labels_order` parameter must be set, it must be one of: Alphabetical, Ascending, Descending, Leaves, None.
```

This makes the SQLite writer **completely unusable** via the standard `BioCypher(dbms="sqlite")` path.

### Steps to reproduce

```python
import biocypher
bc = biocypher.BioCypher(dbms="sqlite")
# ValueError: A batch writer `labels_order` parameter must be set …
```

### Current `sqlite` section in `biocypher_config.yaml`

```yaml
sqlite:
  database_name: sqlite.db
  quote_character: '"'
  delimiter: '\t'
  # labels_order / node_labels_order / edge_labels_order are ABSENT
```

### Fix

Add the missing keys (matching the pattern used by `postgresql`, `neo4j`, `rdf`, and `owl`):

```yaml
sqlite:
  database_name: sqlite.db
  quote_character: '"'
  delimiter: '\t'
  labels_order: "Ascending"
  node_labels_order: "None"
  edge_labels_order: "None"
```

This is the same bug that was fixed for `arangodb` in PR #574.

### Python version

Any

### BioCypher version

current `main`

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

felipebridge:
[...]

PR #516, opened 2026-05-16
feat(translate): deprecate label_as_edge; add synonym_for support for edges

slobentanzer:
## Summary

Closes #435

`synonym_for` already renames nodes in the ontology NX graph — the exact same mechanism works for **edge types** (associations). This PR makes that explicit and deprecates the older `label_as_edge` field in favour of it.

### The problem with `label_as_edge`

When a schema entry uses `label_as_edge: PERTURBED_IN_DISEASE`, the KG output uses `PERTURBED_IN_DISEASE` as the relationship type, but the ontology graph still contains the original node `gene to disease association`. The two are now disconnected: ancestor look-ups on `PERTURBED_IN_DISEASE` fail with a `NetworkXError`, so the `_batch_writer` has to catch that error and fall back to a flat label. This is the exact issue that @jdreo flagged — `label_as_edge` allows a disconnected taxonomy graph.

### The fix — `synonym_for` for edges

```yaml
# Old (deprecated):
gene to disease association:
  represented_as: edge
  label_as_edge: PERTURBED_IN_DISEASE
  input_label: [protein_disease, gene_disease]

# New (preferred):
PERTURBED_IN_DISEASE:
  synonym_for: gene to disease association
  represented_as: edge
  input_label: [protein_disease, gene_disease]
```

With the new form, `_add_properties()` renames the `gene to disease association` node in the NX graph to `PERTURBED_IN_DISEASE`, just as it does for node synonyms (e.g., `complex: synonym_for: macromolecular complex`). The renamed node stays in the graph, so ancestry traversals work correctly.

## Changes

- **`biocypher/_translate.py`**: emit `DeprecationWarning` in `_update_ontology_types()` for every schema entry that uses `label_as_edge`, with an actionable message showing the `synonym_for` equivalent.
- **`biocypher/output/write/_batch_writer.py`**: update the existing `logger.warning` fallback message to also point to `synonym_for`.
- **`biocypher/_config/test_schema_config_edge_synonym.yaml`**: new minimal test schema demonstrating `synonym_for` on an edge type.
- **`test/fixtures/ontology.py`**: add module-scoped `edge_synonym_ontology` fixture using the new schema.
- **`test/test_ontology.py`**: `test_edge_synonym_for_in_ontology` — verifies the biolink class is renamed to the schema key and ancestors are reachable.
- **`test/test_translate.py`**:
  - `test_label_as_edge_deprecation_warning` — mock-based; confirms `DeprecationWarning` is raised when a schema has `label_as_edge`.
[...]

PR #577, opened 2026-06-25
Aggregate batched fix PRs (staging → main)

slobentanzer:
## Aggregated small-fix PRs

This batches **23 independently-reviewed PRs** (authored by the routine under @slobentanzer) onto a `staging` branch, each squash-merged as one conventional commit so `release-please` produces a clean changelog on merge to `main`. No releases were cut while batching (`release-please` only fires on push to `main`).

### Bug fixes (`fix:`)
- #544 get: `get_cached_version` returns file paths for extracted archives
- #545 config: replace deprecated `preferred_id` with `namespace` in test/tutorial schemas
- #546 translate: `_filter_props` must not mutate the shared schema dict in strict mode
- #547 core: replace non-existent `start_ontology()` calls with `_get_translator()`
- #548 neo4j: replace dots with underscores in label names
- #549 ontology: replace fragile `str(None)` sentinel with explicit `has_label` check
- #550 batch_writer: route scalar string quoting through `_quote_string()`
- #552 / #553 / #557 / #570: fix garbled / tuple-ified error messages (translate, core, misc, ontology, batch_writer)
- #554 core: `add_nodes`/`add_edges` crash on first call when `_nodes`/`_edges` is None
- #556 config: `config()` setter raised `KeyError` for keys absent from defaults
- #559 postgresql: missing f-string prefix in `_get_data_type` log
- #560 batch_writer: correct log message key in `edge_labels_order` fallback
- #563 biopathnet: use full Ontology graph so synonym types are found (#489)
- #567 postgresql: only inject `PGPASSWORD` when a password is set
- #569 get_writer: forward postgresql `host` config key as `db_host`
- #571 workflow: keep `_seen_nodes`/`_seen_edges` in sync with graph mutations
- #574 config: add missing `arangodb` section to default config
- #576 config: add missing `labels_order` keys to sqlite section (#575)

### Features (`feat:`, low-risk/additive)
- #565 neo4j: add `import_call_additional_options` config key
- #511 config: warn about unknown top-level keys in `biocypher_config.yaml` (also adds `airr` section)

### Conflict resolutions applied during aggregation
- Test files (`test_postgres.py`, `test_core.py`, `test_translate.py`, `test_config.py`, `test_neo4j.py`) — unioned the test functions added by overlapping PRs.
- `biocypher_config.yaml` — sqlite `labels_order` (#576) placed in the sqlite section; #511's duplicate `arangodb` block dropped in favour of #574's.
- `_config/__init__.py` test interaction — #556's "absent key" tests retargeted from `arangodb` (now a default, courtesy of #574) to a synthetic `custom_unknown_dbms` key.

### Verification
- Local `pytest` over all touched modules: **225 passed, 4 skipped** (Neo4j/Postgres container tests skip locally).
- Closed #562 (superseded by #571 — same four methods, with correct `deduplication` guards).

### Out of scope (left as standalone PRs to `main`)
- #516 (deprecate `label_as_edge`) — rebased + conflict resolved, needs review of the deprecation path.
- #555 (varying-property node header groups) — large refactor, awaiting FHIR-data validation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/biocypher/_config/__init__.py
+++ b/biocypher/_config/__init__.py
@@ -124,7 +124,12 @@ def config(*args, **kwargs) -> Optional[Any]:
         return result[0] if len(result) == 1 else result
 
     for key, value in kwargs.items():
-        globals()["_config"][key].update(value)
+        if value is None:
+            globals()["_config"][key] = None
+        elif key in globals()["_config"] and isinstance(globals()["_config"][key], dict) and isinstance(value, dict):
+            globals()["_config"][key].update(value)
+        else:
+            globals()["_config"][key] = value
 
 
 def reset():
--- a/biocypher/_config/__init__.py
+++ b/biocypher/_config/__init__.py
@@ -81,6 +81,9 @@ def read_config() -> dict:
     # TODO account for .yml?
     local = _read_yaml("biocypher_config.yaml") or _read_yaml("config/biocypher_config.yaml") or {}
 
+    _warn_unknown_keys(local, defaults, "biocypher_config.yaml")
+    _warn_unknown_keys(user, defaults, "user config")
+
     for key in defaults:
         # Apply user-level settings on top of defaults first.
         # An explicit `null` in user config clears the default (e.g.
@@ -106,6 +109,20 @@ def read_config() -> dict:
                 defaults[key] = value
 
     return defaults
+
+
+def _warn_unknown_keys(config_dict: dict, valid_keys: dict, source: str) -> None:
+    """Warn about keys in config_dict that are not recognised in valid_keys."""
+    for key in config_dict:
+        if key not in valid_keys:
+            valid = sorted(k for k in valid_keys if k != "Title")
+            warnings.warn(
+                f"Unknown configuration key '{key}' found in {source}. "
+                f"This may be a misspelling. Valid top-level keys are: "
+                f"{', '.join(valid)}.",
+                category=UserWarning,
+                stacklevel=4,
+            )
 
 
 def config(*args, **kwargs) -> Optional[Any]:
--- a/biocypher/_config/biocypher_config.yaml
+++ b/biocypher/_config/biocypher_config.yaml
@@ -183,6 +183,22 @@ sqlite:
   # import_call_bin_prefix: '' # path to "sqlite3"
   # import_call_file_prefix: '/path/to/files'
 
+arangodb:
+  ### ArangoDB configuration ###
+  database_name: biocypher # DB name
+
+  # ArangoDB import batch writer settings
+
+  delimiter: ","
+  array_delimiter: "|"
+  quote_character: '"'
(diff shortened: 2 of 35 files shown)


slobentanzer:
Hi @ecarrenolozano @iulusoy I packed all small fixes of recent times (incl. workshop suggestions) into this PR; no need for a complete review as it is now an aggregate, but maybe you have some feedback on the behavioural changes. I summarised them below using Claude. In my opinion, nothing is really breaking anything (still will be a minor bump), could cause some backward compatibility issues (the first one in particular was a bit of a mixed decision). But overall, if you don't see a major problem, I'd merge this. Can one of you approve? Thanks!

Worth a deliberate look (behavior/API/output changes)

  - #548 — Neo4j label sanitization. Decides to replace . with _ in label names (e.g. wikipathways.pathway →
  wikipathways_pathway) instead of backtick-escaping. This changes the actual label strings in output for anyone with
  dotted labels. Philosophically: "labels should be identifier-safe" vs "preserve label text verbatim."
  - #550 — Scalar string quoting. Routes scalar string properties through _quote_string(), so embedded quote chars get
  doubled/escaped in CSV output. Correct for import safety, but changes the byte content of output files vs. before.
  - #544 — get_cached_version return semantics. For extracted archives it now walks the directory and returns a list of
  file paths instead of the directory path. Callers that expected a dir get different output — a contract change, not
  just a crash fix.
  - #549 — Unlabeled ontology nodes. Replaces the str(None)=="none" sentinel with an explicit has_label() check and
  silently removes nodes lacking rdfs:label (with a debug log). Decision: drop-and-log rather than raise.
  - #563 — biopathnet uses ontology._nx_graph. Reaches into a private attribute (full graph) instead of the
  head-ontology accessor, to find synonym types. Works, but couples to internals.
  - #511 — Unknown-config-key warning. Emits a UserWarning (not an error) for any top-level key not in defaults.
  Decision: warn, don't fail. Also adds an empty airr: placeholder section. Could be noisy for users who intentionally
  carry extra keys.

  Defaults / config values someone chose

  - #574 — arangodb defaults: database_name: biocypher, quote_character: '"', labels_order: Ascending. (Note: #511 had
  proposed different values — _system, '. I kept #574's and dropped #511's duplicate; flag if you'd prefer #511's.)
  - #576 — sqlite gets labels_order/node/edge: Ascending/None/None.
  - #556 — config setter now creates absent keys and accepts None instead of KeyError/AttributeError. Decision:
  permissive setter.
  - #567 — PGPASSWORD only injected when a password is actually set (not an empty var).
  - #569 — postgres host now forwarded as db_host (was silently dropped).
  - #565 — import_call_additional_options: new optional Neo4j config key, fully additive.
  - #547 — lazy translator: translate_term/translate_query etc. now lazily init the translator via _get_translator() on
  first use.

iulusoy (approved):
It looks good to me, thanks.

PR #578, opened 2026-06-29
fix(biopathnet): get types from assembled ontology + manage all spaces in properties

jdreo:
Types were search in the head taxonomy.
Tabs were not managed from a tab-spaced input.

PR #491, opened 2026-03-18
Fix biopathnet synonyms

clairelaudy:
Fixes the missing synonyms nodes in the hierarchy, when applying _biopathnet writer.

Get the assembled ontology in the _biopathnet.py writer instead of head ontology.

refs: https://github.com/biocypher/biocypher/issues/489

Issue #500, opened 2026-04-15
[FEAT]: GraphQL schema generator for BioCypher


carissableker:
### Feature Type

Add new functionality to BioCypher

### 🚀 The feature and motivation

**Feature description**
As a developer,  I would like the ability to generate a  [Neo4j GraphQL Library](https://neo4j.com/docs/graphql/current/) compatible GraphQL schema directly from a BioCypher schema file. This schema should map BioCypher entities, properties, and relationships to GraphQL types and directives like `@node` and `@relationship`. This would allow me to automatically expose BioCypher knowledge graphs via a standardized API (beyond direct Cypher queries) in sync with the underlying data model. 

**Motivation**
Neo4j includes an [introspection tool](https://neo4j.com/docs/graphql/5/introspector/) that generates a schema from an existing database. However, this approach has limitations, including missing ontological hierarchy. Otherwise manual work is needed to maintain consistency  between the BioCypher and GraphQL schemas. A direct tool will also ensure that GraphQL types match BioCypher generated class names.

**Example** 
(incomplete, just for illustration)

BioCypher configuration for a protein-protein interaction:

```yaml
protein:
  represented_as: node
  preferred_id: uniprot
  properties:
    name: str

protein protein interaction:
  is_a: pairwise molecular interaction
  represented_as: edge
  source: protein
  target: protein
  properties:
    score: float
```

GraphQL schema for the Neo4j GraphQL Library:

``` .graphql
type Protein @node(labels: ["Protein"]) {
  uniprotId: ID!
  name: String
  interactsWith: [Protein!]! @relationship(type: "PROTEIN_PROTEIN_INTERACTION", direction: OUT, properties: "InteractsWithProps")
}

interface InteractsWithProps @relationshipProperties {
  score: Float
}
```

### Additional context

I do not need this feature anytime soon, but I anticipate needing it in the future and am opening this issue now in case there is interest from others.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

slobentanzer:
Congrats on opening issue #500 🎉

Sounds reasonable; maybe consider for the June workshop? I'll add a specific tag for the workshop and go through other issues. We should do a clean-up anyway.

jdreo:
This is similar to PR #455 (a backend for exporting a metagraph through NetworkX) which is waiting for issue #435 (get rid of `label_as_edge`) to be solved before going forward.

slobentanzer:
[...]

Issue #435, opened 2025-04-14
[FEAT]:  Harmonise re-labelling: edges should also be subject to the synonym process


slobentanzer:
### Feature Type

Add new functionality to BioCypher

### 🚀 The feature and motivation

For nodes, we currently have the "synonym" feature to translate labels (e.g., because of readability issues) into other labels for use in the KG. For edges, the `label_as_edge` is used for historical reasons. This is potentially confusing to new users. Solution: we should update the schema configuration to harmonise the synonym process, making it accessible for edges as well, and deprecating the `label_as_edge`.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Shashwat26Bhatnagar:
I can work on this along with one of my friends.
Can you please assign this task to us?

mesinghaman:
Yeah we can work on it.

slobentanzer:
Fully agree, thanks for the explanation @jdreo

PR #580, opened 2026-07-10
Biopathnet integration

jdreo:
This PR integrates and supersedes PRs #491 and #578.

It finalizes the BioPathNet backend, adds doc and adds test.

It also clarifies one error message in deduplicate.


slobentanzer:
Hi @jdreo, many thanks; CI does not pass. Did you also check #435; as that was the blocker for you? (See conversation in #500.)

PR #584, opened 2026-08-05
fix(neo4j)!: match neo4j-admin's own delimiter defaults

### LINKED ISSUE #583: Neo4j CSV import fails on Neo4j 2026.06+: default delimiter ';' clashes with --vector-delimiter
## Summary

BioCypher's default Neo4j CSV output cannot be imported on Neo4j **2026.06.0** or later. `neo4j-admin` aborts during argument validation:

```
Character ';' specified by vector delimiter is the same as specified by delimiter
```

Neo4j added a `--vector-delimiter` option (*"Delimiter character between vector coordinates within a value in CSV data"*) whose default is `;`, and it now refuses to run when that equals `--delimiter`. BioCypher's packaged default is `delimiter: ";"`, so the generated `neo4j-admin-import-call.sh` hits the clash out of the box.

This is independent of #522 / Parquet, and it affects the CSV path specifically, which is what we recommend as the fallback for older Neo4j.

## Affected versions

Measured against the generated import call, with BioCypher's default `delimiter: ";"`:

| Neo4j | CSV import |
| --- | --- |
| 5.26.28 (LTS) | ok — no `--vector-delimiter` option exists |
| 2026.04.0 | ok — option exists, clash not enforced |
| 2026.05.0 | ok — option exists, clash not enforced |
| **2026.06.0** | **fails** |

Parquet output is **not** affected: with `--input-type=parquet` the same `--delimiter=";"` is accepted and the import succeeds. The validation appears to apply to CSV input only.

## Reproduce

```bash
# write any graph with the Neo4j writer and file_format: csv, then:
docker run --rm -v "<biocypher-out-dir>:/import:ro" --entrypoint bash neo4j:2026.06.0-community -c \
  'neo4j-admin database import full neo4j \
     --delimiter=";" --array-delimiter="|" --quote="'"'"'" \
     --overwrite-destination=true \
     --nodes="/import/Protein-header.csv,/import/Protein-part.*"'
```

Adding `--vector-delimiter=","` (any character other than the field delimiter) makes the same call succeed.

## Why this is not a one-line fix

`--vector-delimiter` does not exist on the 5.26 LTS line — `neo4j-admin database import full --help` on 5.26.28 has no such option — so emitting it unconditionally would break every LTS user with `Unknown option`. This is the same shape as the `--input-type` problem discussed in #522: the target Neo4j version is only known at import time, not while BioCypher writes the files.

Options, roughly in order of preference:

1. **Emit `--vector-delimiter` conditionally**, from the same version check that the generated script already performs. The script branches on major version today (`neo4j-admin --version | cut -d '.' -f 1`), which is too coarse — it cannot distinguish 2026.05 from 2026.06, or 5.26 from a calendar release. It would need to compare the full version.
2. **Change the packaged default `delimiter`** away from `;`. Fixes every version at once and needs no detection, but it is a breaking change for anyone whose downstream tooling reads the CSV output, and it silently changes existing pipelines' output.
3. **Document it** and let users set a non-clashing `delimiter` themselves. Cheapest, but the failure only surfaces at import time, after a potentially long build.

Worth noting that a general mechanism for "emit this flag only for Neo4j versions in range X" would cover both this and `--input-type`, rather than solving each one separately.

## How this was found

By the end-to-end import test added in #522, which runs the generated call against a real `neo4j-admin` in a container. Nothing in the suite executed the generated script before that.

slobentanzer:
Fixes #583.

## Problem

BioCypher defaults the Neo4j CSV output to `;` field delimiter, `|` array delimiter and `'` quote. Neo4j 2026.06 added `--vector-delimiter`, defaulting to `;`, and refuses to run when it equals `--delimiter`:

```
Character ';' specified by vector delimiter is the same as specified by delimiter
```

The generated `neo4j-admin-import-call.sh` therefore fails on Neo4j 2026.06+ when `file_format: csv` is set. The default Parquet output is not affected: its import call omits the delimiter and quote flags entirely. See #583 for the version-by-version measurements.

## Change

Adopt `neo4j-admin`'s own defaults:

| | before | after | `neo4j-admin` default |
| --- | --- | --- | --- |
| `delimiter` | `;` | `,` | `,` |
| `array_delimiter` | `\|` | `;` | `;` |
| `quote_character` | `'` | `"` | `"` |

The previous values were picked to be less likely to collide with the data. That solved a problem Neo4j owns, and bought a collision with a flag that did not exist yet. Following Neo4j's defaults means new import options cannot clash with us by construction: Neo4j's own array and vector delimiters are both `;`, and that is *not* rejected, because only field-vs-vector is validated.

If your data contains the delimiter, the answer is Neo4j's quoting, or the default Parquet output (#522), rather than BioCypher choosing rarer characters.

## Verification

Generated CSV output (`file_format: csv`, otherwise the shipped default config), then ran the generated import call against real `neo4j-admin` in containers:

| image | result |
| --- | --- |
| `neo4j:2026.06.0-community` | `IMPORT DONE` (was: vector delimiter clash) |
| `neo4j:5.26.28-community` | `IMPORT DONE` |
| `neo4j:4.4-community` | `IMPORT DONE` (Neo4j 4 form of the call) |

Full suite after the rebase onto main: 479 passed, 27 skipped.

## Breaking change

The Neo4j CSV output (`file_format: csv`) now uses `,` / `;` / `"` instead of `;` / `|` / `'`. Anything downstream that parses BioCypher CSV output needs updating. Existing behaviour is one config block away:

```yaml
neo4j:
  delimiter: ";"
  array_delimiter: "|"
  quote_character: "'"
```

After the rebase onto main, no test changes remain in this PR: the integration tests now read Parquet, where no delimiters apply, and the CSV writer tests set their delimiters explicitly.

## Not included

Emitting no delimiter flags at all when they match Neo4j's defaults. That would track whatever the target version defaults to, which is tempting, but BioCypher writes the files, so the two must agree. Passing them explicitly is what keeps the written output and the import call from drifting apart if Neo4j ever changes a default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/biocypher/_config/biocypher_config.yaml
+++ b/biocypher/_config/biocypher_config.yaml
@@ -96,9 +96,15 @@ neo4j:
 
   ## Neo4j admin import batch writer settings
 
-  delimiter: ";"
-  array_delimiter: "|"
-  quote_character: "'"
+  # These match neo4j-admin's own defaults. Keep them in step with Neo4j
+  # rather than picking characters that look less likely to occur in the data:
+  # BioCypher's previous ";" field delimiter collided with the
+  # --vector-delimiter default that Neo4j added in 2026.06, which makes the
+  # generated import call fail outright.
+  # See https://github.com/biocypher/biocypher/issues/583
+  delimiter: ","
+  array_delimiter: ";"
+  quote_character: '"'
 
   # File format of the node and edge data files. Headers are always CSV.
   # `parquet` requires PyArrow (installed with `biocypher[neo4j]`) and a target
--- a/biocypher/_config/biocypher_config.yaml
+++ b/biocypher/_config/biocypher_config.yaml
@@ -102,14 +102,14 @@ neo4j:
   # --vector-delimiter default that Neo4j added in 2026.06, which makes the
   # generated import call fail outright.
   # See https://github.com/biocypher/biocypher/issues/583
-  delimiter: ","
-  array_delimiter: ";"
-  quote_character: '"'
+  csv_column_delimiter: ","
+  csv_array_delimiter: ";"
+  csv_string_quote_character: '"'
 
   # File format of the node and edge data files. Headers are always CSV.
   # `parquet` requires PyArrow (installed with `biocypher[neo4j]`) and a target
-  # Neo4j version that supports import with Parquet; the delimiter, array
-  # delimiter and quote character above apply to the `csv` data files only.
+  # Neo4j version that supports import with Parquet; the CSV options above
+  # apply to the `csv` data files only.
   file_format: parquet # or `csv`
 
   # How to write the labels in the export files.
--- a/biocypher/output/write/_get_writer.py
+++ b/biocypher/output/write/_get_writer.py
@@ -111,9 +111,9 @@ def get_writer(
         return writer(
             translator=translator,
             deduplicator=deduplicator,
-            delimiter=dbms_config.get("delimiter"),
-            array_delimiter=dbms_config.get("array_delimiter"),
-            quote=dbms_config.get("quote_character"),
+            delimiter=dbms_config.get("csv_column_delimiter") or dbms_config.get("delimiter"),
+            array_delimiter=dbms_config.get("csv_array_delimiter") or dbms_config.get("array_delimiter"),
+            quote=dbms_config.get("csv_string_quote_character") or dbms_config.get("quote_character"),
             output_directory=output_directory,
             db_name=dbms_config.get("database_name"),
             import_call_bin_prefix=dbms_config.get("import_call_bin_prefix"),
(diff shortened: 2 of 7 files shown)


kpto (dismissed):

slobentanzer:
@kpto had to resolve conflicts with main here, still good to go? I'd merge this and then bump minor.

kpto:
@slobentanzer I’m wondering whether it would be better to rename those parameters to:
```YAML
neo4j:
  csv_column_delimiter: ";"
  csv_array_delimiter: "|"
  csv_string_quote_character: "'"
```

while keeping the old parameters available as aliases for backward compatibility, but removing all references to them from the documentation.

What do you think?

slobentanzer:
@kpto good idea, agree

kpto:
@slobentanzer I’ve made the changes, please take a look.

slobentanzer:
@kpto looks good, thanks. I can't approve my own PR though.

kpto (approved):
I approve my changes 👍

PR #591, opened 2026-09-14
Reorganization of the documentation - tutorials section and API description

iulusoy:
This PR restructures some of the biocypher documentation, in particular with respect to the tutorials:

**Learn section**
  - New three-way split: Tutorials / HowTo-FAQ / Explanations, each with its own index.md overview (grid-card layout for Tutorials and Explanations; HowTo, still under construction)
  - New `docs/learn/index.md` top-level landing page
  - New `docs/learn/tutorials/catalog.md` - tags for a difficulty-driven tutorial catalog, generated by a new mkdocs hook (`docs/hooks/docs_hooks.py` + `test/docs/hooks/test_docs_hooks.py`)

**New tutorials**
  - OntoWeaver tutorial (`tutorial_ontoweaver/`, incl. SVG/PNG assets)
  - Agentic/LLM-assisted basics tutorial (`tutorial_basic_with_agents/`, incl. ~15 screenshot assets)
  - Existing tutorials (001–003, Neo4j offline) reworked/expanded, `agent_api_guide.md` updated

 **New explanation pages**
  - `basics-knowledge-graphs.md`, `knowledge-graph-biocypher.md`, `schema-config-philosophy.md`, `graph_class_guide.md`, `architecture-migration.md`, `ontologies.md`, `adapters.md` - all new or substantially rewritten, partially from upstream
  - `about-documentation.md` removed (superseded by the new overview pages)

**New guides**
  - large-scale-builds.md, standalone Docker guide taken from upstream

  **Site/tooling** - partially from upstream
  - `mkdocs.nav.yml` and `mkdocs.yml` reorganized to match new structure
  - `docs/stylesheets/extra.css` added
  - `docs/api.rst` → `docs/api.md` (mkdocstrings format switch)
  - `docs/llms*.txt` moved under new `docs/agents/` dir; `docs/llms.md` updated
  - `.pre-commit-config.yaml` updated
  
These changes represent the main restructure, now under this new structure we will expand the how-to and the tutorials a bit - part of the tutorials (ie the descriptions about neo4j) will be moved to how-tos. The old python script-like tutorials, that are incompatible with windows are planned for removal.

PR #592, opened 2026-09-14
ci: check docstrings against signatures with pydoclint

### LINKED ISSUE #354: Automatically ensure, that docstrings are up to date
Just realized, that sometimes the docstrings tend to become outdated, without that being noticeable (e.g. one parameter added to a function and docstring not adapted).
So it would be great to automatically check. if the docstrings are up to date and fit together with the method/class declaration.

felipebridge:
## Summary

Closes #354.

Adds [pydoclint](https://github.com/jsh9/pydoclint) (the tool @nilskre suggested in the issue thread) as a pre-commit hook, so a docstring that goes out of sync with its function's parameters or return type — the original problem described in the issue — gets caught automatically instead of silently rotting.

## Scope

Running `pydoclint --style=google` over the whole `biocypher/` package today reports violations in 25 of 42 modules (mostly missing `Args:`/`Returns:` sections and mismatched types on existing docstrings, unrelated to this change). Enabling the hook unscoped would make this PR fail CI on pre-existing issues across the codebase, and fixing all of them here would turn a docstring-tooling PR into a huge, unrelated docstring-rewriting PR.

So, same as this repo's own `ruff --select=ALL` (only `E,F,I` are actually enforced in `.pre-commit-config.yaml`, with a `# TODO: enable ruff once it does not throw thousands of errors` note in `.github/actions/test/action.yaml`), I scoped the new hook to the 17 modules that are already clean today via an explicit `files:` allowlist, verified locally file-by-file. The remaining 25 are listed in a comment right below it as a checklist — move a file's regex line up once its docstrings are fixed, and the hook starts protecting it.

This means the hook is fully enforced (not just added-and-ignored) from day one, just on a smaller, honest surface area, with a clear, incremental path to full coverage.

## Changes

- `.pre-commit-config.yaml`: new `pydoclint` hook (pinned to `0.9.1`), scoped via `files:` to the 17 currently-compliant modules, with the non-compliant ones listed in a comment.
- `pyproject.toml` / `uv.lock`: added `pydoclint>=0.9.1` to the `dev` dependency group so it's usable directly (`uv run pydoclint --style=google <path>`) outside of pre-commit too.

## Test plan

- [x] `uv run pre-commit run pydoclint --all-files` — passes
- [x] `uv run pre-commit run pydoclint --files biocypher/_core.py` — correctly skipped (not yet compliant, excluded from scope)
- [x] `uv run pre-commit run pydoclint --files biocypher/_metadata.py` — correctly runs and passes (in scope)
- [x] `uv run pre-commit run --all-files` (full suite: ruff, ruff-format, pydoclint, and the existing hygiene hooks) — all pass, no regressions


slobentanzer:
@felipebridge many thanks! holding for a little while because we want to merge #591 first.

PR #593, opened 2026-09-15
docs: document head_ontology: null (headless mode) in config reference

### LINKED ISSUE #586: [DOC]: headless mode not in doc
### Location of the documentation

The `head_ontology: null` configuration is not explained in the documentation.

### Documentation problem

I was expecting to find it mentioned in the [biocypher-config](https://biocypher.org/BioCypher/reference/biocypher-config/) section.

### Suggested fix for documentation

Mention `head_ontology: null` in the reference table, and add a paragraph explaining it.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

felipebridge:
Fixes #586.

## What

`head_ontology: null` (or omitting `head_ontology` entirely) opts BioCypher into headless mode: no OWL/TTL fetch, no `rdflib` graph, schema validation still runs against `schema_config.yaml` as usual (see `NullOntology` in `biocypher/_ontology.py` and `test/test_headless.py`). The config reference documented `head_ontology.url` / `.root_node` / `.switch_label_and_id`, but never mentioned that `null` is a valid, meaningful value.

## Changes

- `docs/reference/biocypher-config.md`: adds a `head_ontology` row to the parameter table noting the `null` case, and a "Headless mode (no ontology)" subsection explaining what headless mode does and does not support.

## What I checked before writing this

Read `NullOntology` and `BioCypherCore._get_ontology()` in `_ontology.py`/`_core.py` to confirm the actual behavior rather than going off the class docstring alone — the docstring claims the Neo4j connector's "schema dump" raises in headless mode, but `_neo4j_driver.py` only calls `ontology.get_dict()`, which `NullOntology` implements and `test_null_ontology_get_dict_shape` asserts mirrors the real `Ontology.get_dict()` — so that path actually works fine and I left it out of the "cannot do" list. I also checked that `biopathnet` isn't guarded by a `NotImplementedError` the way `show_ontology_structure`/`get_rdf_graph` are — it reads `ontology._nx_graph` directly, which `NullOntology` doesn't define, so I described it as erroring rather than claiming a specific exception type.

No code changes; docs only.


slobentanzer:
@felipebridge thanks; same comment as #592, waiting for #591 merge.

PR #594, opened 2026-09-15
feat(get): load Resource definitions from a YAML config file

### LINKED ISSUE #408: Add option to load resource download definition from config
In order to increase compatibility with CLI-driven workflows, a definition of a `Resource` URL and lifetime in a configuration file, as opposed to the Python code, would be convenient.

Target: create a PR that introduces the option to load `Resource` definitions from file.

felipebridge:
Fixes #408.

## What

Adds `Resource.from_config(path)`, a static factory that loads a list of `FileDownload`/`APIRequest` resource definitions from a YAML config file instead of requiring them to be constructed in Python code — useful for CLI-driven workflows, per the issue.

```yaml
resources:
  - name: uniprot_api
    type: api
    url_s: https://rest.uniprot.org/uniprotkb/P12345.json
    lifetime: 7
  - name: my_download
    type: file
    url_s:
      - https://example.com/a.csv
      - https://example.com/b.csv
    lifetime: 0
    is_dir: false
```

The list may also sit at the top level of the file (no `resources` key) for a config file dedicated purely to resources. `type` selects `FileDownload` (default) or `APIRequest`; every other key is passed straight through as a constructor kwarg, so unknown/misplaced keys (e.g. `is_dir` on an `api` entry) surface as an ordinary `TypeError` from the constructor rather than being silently dropped.

## Why this shape

- Reuses `biocypher._config._read_yaml` (and its `MyLoader`, which warns on double-quote escape-sequence footguns) instead of adding a second YAML-loading code path.
- Kept as a `Resource` classmethod rather than a free function so it reads naturally next to `Downloader.download(*resources)` — `Downloader.download(*Resource.from_config("resources.yaml"))`.

## Tests

Added to `test/test_get.py`, covering: top-level list, nested under `resources:`, multi-URL `url_s`, missing file, missing required key, unknown `type`, and a config that isn't a list.

`python -m pytest test/test_get.py` — 27 passed, 1 skipped (pre-existing, unrelated FTP test).
`ruff check --select=E,F,I` (the rule set the repo's pre-commit hook actually enforces) and `ruff format --check` — both clean.


slobentanzer:
@felipebridge thanks, very nice contribution. Will review soon (a bit swamped right now).

